PR #3, opened 2023-07-02
Adding vehicle reequipping and reworked equipping script

FluffyRaKu:
All vehicles can now be reequipped, although the double-weapon bug currently prevents rhinos being modified.

Equip script now checks the unit type for lists of valid equipment under more circumstances. Dreadnoughts should now only ever see dreadnought gear, predators see pred gear etc. Some vehicles have different weapon slots between slot 1 and slot 2 to differentiate between turrets and sponsons.


ZAB909:
Merging this on my phone, I think it's alright to merge

PR #5, opened 2023-07-14
Event fixes

hristokn:
I have changed the scr_random_event and also added some minor help scripts.

The minor scripts are:
-2 scripts to check if the chapter has some advantage or disadvantage
-some functions that check if a star has a planet with a specific owner\type\feature
-a function that returns all the stars in an array
-change the scr_random_find to be more readable
-a script that converts a company number to a string
-a script that counts how many marines are on a ship
-added some enums to 'macros'

I have tried to improve the scr_random_event script. The script is called at the end of every turn, it decides if a random event should happen, then chooses which event it should happen and then executes the event. 
I think I have broken the 'buy eldar advice' diplomacy action, so fixing that is on the to do list.


ZAB909 (approved):

ZAB909:
eldar diplo is pretty low priority, so all good. make an issue report on this github and link this so we know why it doesn't work, for the future, so we don't forget

PR #13, opened 2023-07-19
Planet feature refurb

OH296:
Still need help on a couple of bits if i don't get around to them:
1.saving and loading of new planet features format
2. for some reason names of features not appearing in planet overviews however all functionality is present purely visual


OH296:
should all be fixed i will make a little doc for the new way to reference planet features

PR #33, opened 2023-08-07
Marine auto load

OH296:
- re-written the scr_start_load script
- refactored (slightly) the unit size script
- swapped all uses of armor to armour because it was pissing me off


OH296:
@GalacticChimp i've tried to actually leave some comments in the scr_start_load to try and make it as intuitive. Let me know if it seems intuitive to you or not as i'm just really trying to make everything as clean and intuitive as i can.

objects/obj_controller/Create_0.gml
```diff
@@ -529,7 +529,7 @@ exit_all=0;
 // 
 diplomacy=0;
 trading=0;
-trading_artifact=0;
+trading_artifat=0;
```
GalacticChimp, line 532: this one was fine

OH296: yeah sorted good catch not sure how that happened

GalacticChimp (changes requested):
Looks fine, just please fix conflicts and retest

PR #40, opened 2023-08-12
Marine rpg

OH296:
-the only thing important to look at here is line 276 onwards or scr_add_man. It outline one method for moving marines into a fully rpg stat based scenario without an entire code rewrite.
It would allow us to greatly improve efficiency by allowing functions to do basic tasks such as changing marine weaponry or a marines location, to far more complex things like giving distinct personalities and talents (with talent trees) to marines, while only having to use a single 2d array to access all of the marine's attributes. the struct is also highly amendable in the instance that we wanted to edit or add new elements to a marines design. 
Particularly look at line 351 and 333 and how the code works much better in conjunction to keep the information of a marine in one place and concise.
While i need to put a bit more though into this i think it could also be used as a way of slowly swapping out the old marine attributes, and eventually fully deprecating the old 2d marine array system in a piecemeal fashion (e.g deprecate the old method with modern varibales stored in the struct in a script by script basis as the old system will still work in parallel) I can do further explanation on this if people would require.
of note is the fact that  this is only a simply out line and the final method would involve a better level of inheritance to differentiate between different types of marines (and all units in the game in general).
I've not spent a lot of time on this as obviously this may not be a direction people want to go but in my eyes it is the obvious easiest and safest way forward.
Overall i think this could save literally thousands of lines of code pretty quickly and pave the way for immenant expansion with elements that are currrently being talked about ina wishful thinking frame of mind.


scripts/scr_add_man/scr_add_man.gml
```diff
@@ -1,47 +1,184 @@
...
+				location_name = obj_ini.loc[company,marine_number]; //system marine is in
+			} else {
+				location_type = "ship"; //marine is on ship
+				location_id = obj_ini.lid[_company, _unit]; //ship array location
+				location_name = obj_ini.ship_location[location_id]; //location of ship
+			}
+			return [location_type,location_id ,location_name];
+		};
+		
+		static load_marine =function(ship){
```
OH296, line 114: @GalacticChimp the standardised struct method for loading a marine to a ship either from a planet or from another ship (will handle wither) will be callable via <<marine>>.load_marine(<<ship>>).

Ignore the fact that all this is being contained in the scr_add_man script, i just prefer to work in  existing files and then split the scripts up later....Idk why.

PR #45, opened 2023-08-15
Marine auto load

OH296:
-rewritten the scr_start_load script to function properly, technically no longer ignores escort but has a new algorithm that makes all endeavours to sort onto non escort ships.
Will edit more to reintroduce escort options and to add options to distribute specialists equally on ships but for now this makes it actually work compared to current state


objects/obj_controller/Create_0.gml
```diff
@@ -529,7 +529,7 @@ exit_all=0;
 // 
 diplomacy=0;
 trading=0;
-trading_artifact=0;
+trading_artifat=0;
```
GalacticChimp, line 532: trading_artifa**c**t=false

scripts/scr_unit_size/scr_unit_size.gml
```diff
@@ -1,23 +1,28 @@
...
     // role: role
     // other_factors: +?
+	//mobility item
     var sz = 1;
     if (role!=""){
-
-        if (string_count("Dread",armour)>0) {sz=5;}
-        else if (role="Rhino") {sz=10;}
+		var bulky_armour = ["Terminator Armour", "Tartaros"]
+        if (string_count("Dread",armour)>0) {sz+=5;}else if (array_contains(bulky_armour,armour)){sz +=1};
```
GalacticChimp, line 12: please move the else if onto it's own line, too hard to read

objects/obj_ground_teams/Alarm_0.gml
```diff
@@ -0,0 +1,31 @@
...
+            if (feature.ruins_race=6) then newline+="Eldar colonization structures from an unknown time.";
+            if (feature.ruins_race=10) then newline+="golden-age Imperial ruins, since decorated with spikes and bones."; 
+			scr_newtext();
+	}else{			
+		newline = "The ruins seem much unchange from the last exploration records"
+		scr_newtext();
+	}
+	
+	var pathway = choose(1,2,3);
+	if (pathway >0 ){
```
GalacticChimp, line 20: this will always be true, because `var pathway = choose(1,2,3);` will always give at least 1

scripts/scr_start_load/scr_start_load.gml
```diff
@@ -1,547 +1,121 @@
-function scr_start_load(argument0, argument1, argument2) {
-    // argument0: the fleet object
-    // argument1: star object
-    // argument2: 1 for including escorts, 2 for no escorts
+function scr_start_load(fleet, load_from_star, escort_load) {
+    // fleet: the fleet object
+    // load_from_star: star object
+    // escort_load: 1 for including escorts, 2 for no escorts
```
GalacticChimp, line 4: 👍

GalacticChimp (changes requested):
Added some comments. I'll also pull and test this later today.

objects/obj_ground_teams/Alarm_0.gml
OH296: was done for testing purposes and will be overwritten in a different in a future pr

OH296: actually not entirely sure how this has ended up in this branch... one of lives mysteries

OH296:
@GalacticChimp thanks for looking over. All changes implemented.

scripts/scr_unit_size/scr_unit_size.gml
```diff
@@ -1,23 +1,30 @@
...
+
+// lets face it this should all be solved in a method in the marines struct....
+function scr_unit_size(armour, role, other_factors, mobility) {
 
     // armour: armour
     // role: role
     // other_factors: +?
+	//mobility item
     var sz = 1;
     if (role!=""){
```
GalacticChimp, line 10: you can simplify this, by inverting this if and making it a guard clause:

```
if(role == "") then return 0; 
...
```
it will make reading easier and simplify the flow a bit

GalacticChimp, line 10: you can simplify this, by inverting this if and making it a guard clause:

```
if(role == "") then return 0; 
...
```
it will make reading easier and simplify the flow a bit

scripts/scr_unit_size/scr_unit_size.gml
```diff
@@ -1,23 +1,30 @@
...
-
-        if (string_count("Dread",armour)>0) {sz=5;}
-        else if (role="Rhino") {sz=10;}
+		var bulky_armour = ["Terminator Armour", "Tartaros"]
```
GalacticChimp, line 11: `;`

scripts/scr_unit_size/scr_unit_size.gml
```diff
@@ -1,23 +1,30 @@
...
-function scr_unit_size(armour, role, other_factors) {
+
+// lets face it this should all be solved in a method in the marines struct....
+function scr_unit_size(armour, role, other_factors, mobility) {
 
     // armour: armour
     // role: role
     // other_factors: +?
+	//mobility item
     var sz = 1;
```
GalacticChimp, line 9: sz? could you rename this?

scripts/scr_unit_size/scr_unit_size.gml
```diff
@@ -1,23 +1,30 @@
...
     if (role!=""){
-
-        if (string_count("Dread",armour)>0) {sz=5;}
-        else if (role="Rhino") {sz=10;}
+		var bulky_armour = ["Terminator Armour", "Tartaros"]
+        if (string_count("Dread",armour)>0) {
+		sz+=5;
+	}else if (array_contains(bulky_armour,armour)){sz +=1};
+        if (role="Rhino") {sz=10;}
```
GalacticChimp, line 15: the above 3 ifs are very hard to read, please change to something like:

```
        if (string_count("Dread",armour)>0) { sz+=5;}
	else if (array_contains(bulky_armour,armour)) {sz +=1;};
	// newline, because we are starting a new if chain below
        if (role="Rhino") {sz=10;}
```

scripts/scr_unit_size/scr_unit_size.gml
```diff
@@ -1,23 +1,30 @@
...
         else if (role="Predator") {sz=10;}
         else if (role="Land Raider") {sz=20;} 
         else if (role="Land Speeder") {sz=5;}
         else if (role="Whirlwind") {sz=10;}
         else if (role="Harlequin Troupe") {sz=5;}
+		else if (role="Chapter Master"){sz+=1;}
```
GalacticChimp, line 21: you can use `sz++` instead of `sz+=1`, but it's a style issue basically, your choice

scripts/scr_unit_size/scr_unit_size.gml
```diff
@@ -1,23 +1,30 @@
...
     
         if (other_factors=false) then sz=sz*-1;
```
GalacticChimp, line 23: you can use `sz *= -1` instead of `sz = sz*-1`

scripts/scr_start_load/scr_start_load.gml
```diff
@@ -1,547 +1,122 @@
...
-    battle_barges
-    strike_cruisers
-    gladius
-    hunters
-    */
-
-    var splinter, _companies, _equip, remove_size, company_size, ship, ship_size, companies_loaded;
+	
+	
+	// i feel like there definatly is or should be a generic function for this????
```
GalacticChimp, line 9: in an OOP language this would be trivial to do, but here we are with gml2.

the only thing I can think of right now, is map vehicles and marines into a common, simple data structure and do the ship loading operation on that simpler object. But this will probably require a big rewrite

scripts/scr_start_load/scr_start_load.gml
```diff
@@ -1,547 +1,122 @@
...
-            obj_ini.loc[_companies, _equip] = obj_ini.ship_location[ship];
-            company_size += 1;
+	
+		//loop through companies. try and load whole company onto single ship else spread company across largest ships with remaining space
+	for (var _comp =0; _comp<10;_comp++;){
+		var _company_size = 0;
+		var company_loader =[];//array of companies marines
+		var company_vehicle = [];//array of companies vehicles
+		var  ship_fit = true;
+		for (var _marine =1;_marine<(array_length(obj_ini.role[_comp])-1);_marine++){
```
GalacticChimp, line 42: var is named `_marine`, but below it's used in vehicle context: `if (_marine <array_length(obj_ini.veh_role[_comp]))`. I'd recommend to change `_marine` into `_company_object_index` or `_company_asset_index` or something. To make it a bit clearer.

scripts/scr_start_load/scr_start_load.gml
```diff
@@ -1,547 +1,122 @@
...
+							   ship_has_space = false;
+							   break;
+						  }
+					  }					  
+					   company_loader= comp_edit;
+					   company_vehicle = veh_edit;
+				 }
+				 }
+			 }
+		 } else{
```
GalacticChimp, line 113: this is case when a single ship can fit all guys?

GalacticChimp (changes requested):
Added some more comments

scripts/scr_unit_size/scr_unit_size.gml
OH296: honestly think i'm gonna redo all of this file as it's pretty much all copied from what was there, and i'm not happy with it at all.

scripts/scr_unit_size/scr_unit_size.gml
OH296: again gonna rewrite file

scripts/scr_start_load/scr_start_load.gml
OH296: well i've know made a generic method set for moving marines within the marine struct i've been working on so that should solve this (sort of)...

scripts/scr_start_load/scr_start_load.gml
OH296: yeah that's the one could be reordered to go before other stuff if more readable?

scripts/scr_start_load/scr_start_load.gml
OH296: oversight from lazy copy and pasting, will fix.

scripts/scr_start_load/scr_start_load.gml
GalacticChimp: usually we order if-elses by probability of happening, unless the if-else is simple, then go for readability. This one is just a simple bool check, so it can stay, unless you prefer to order. But it doesn't really matter in this case.

scripts/scr_unit_size/scr_unit_size.gml
GalacticChimp: sure, if you're not satisfied with the code, then by all means do a redo

scripts/scr_start_load/scr_start_load.gml
GalacticChimp: could you point me towards that method?

OH296: made a comment in the correct place in marine_rpg pr

GalacticChimp:
converted to draft, as you want to redo this one

PR #52, opened 2023-08-25
[made by Frost] disposition system changes

GalacticChimp:
[Provided by @death0frost] The new disposition system is intricately tied to your influence. As you visit planets, they become integrated into your sphere of control. Within these sectors, factions will request your aid in safeguarding their key locations. For instance, the Adeptus Mechanicus values their Forge Worlds, while the Sisters of Battle prioritize shrines and cathedrals. Allowing these sites to be overwhelmed will harm your reputation and influence as your overall faction standing diminishes due to your perceived negligence.
This can be distilled into the following categories:

1- Imperium: As the commander of the Imperial sector, your efforts to secure planets for them won't be rewarded with disposition points. The scale of planetary management makes it nearly impossible to monitor every world, suddenly the term "needle in a haystack." is relevant. You'll gain 10 disposition points for successfully safeguarding an Imperium-controlled planet, but you'll experience a deduction of 10 points if the planet falls to enemy forces.

Sisters of Battle: The Sisters of Battle have a distinct approach. Their influence surpasses that of the Mechanicus, so allowing numerous sacred sites to be lost will displease their leader. A loss of 5 disposition points will result from such failures. Just like with Imperium planets, securing a planet for the Sisters of Battle will grant you 10 disposition points, while losing the planet will subtract 10 points. Note: You cannot politely ask the sister CEO to resign (you cannot assassinate shrine world governor)

Adeptus Mechanicus: Forge worlds are precious and not numerous, you better have a very good reason if you allow their fall kid. Permitting their overrun will lead to a significant faction-wide reduction of 10 disposition points, as well as 10 points for the specific planet. Carrot in this case is the support of the Mechanicus.

### DIFF
--- a/scripts/scr_enemy_ai_a/scr_enemy_ai_a.gml
+++ b/scripts/scr_enemy_ai_a/scr_enemy_ai_a.gml
@@ -616,14 +616,25 @@ function scr_enemy_ai_a() {
 	                    badd=2;
 	                }
                 
-	                if (owner<=5){
-	                    if (badd=1) and (p_tyranids[run]=0) and (p_necrons[run]=0) and (p_sisters[run]=0) then scr_alert("red","owner",string(name)+" "+string(run)+" has been overwhelmed by Orks!",x,y);
-	                    if (badd=2) and (p_tyranids[run]=0) and (p_necrons[run]=0) and (p_sisters[run]=0){
-	                        scr_popup("System Lost","The "+string(name)+" system has been ovewhelmed by Orks!","orks","");
-	                        scr_event_log("red","System "+string(name)+" has been overwhelmed by Orkz.");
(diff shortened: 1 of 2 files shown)


ZAB909:
Pretty good idea, I think imperium should probably be a lot less. Due to the fact that a player can start with 20-40. And realistically at the start of the game tau will take at least 2, orks will probably make you renegade almost immediately with how much they will take in a short period as well. 

Balancing is very important for this cool feature, otherwise the player will go down a slippery slope. We should figure out a reasonable timer that the player can get a planet back and not have any negatives, but if it times out, then they get pissed?

GalacticChimp:
this will have to wait until the hotfix release is released

GalacticChimp:
Continued in PR #166

PR #79, opened 2023-09-11
Last min changes

ZAB909:
Adds old_guard var to armours for marines, giving them the possibility of being lucky enough to have fancier armour than they should

edited starting armour for command roles

Some commenting and minor bug fixes and additions


ZAB909:
Forgot to mention that corvus and heresy have been buffed damage wise but worsened ac wise

PR #80, opened 2023-09-11
Bump version to 0.8.0.1 + version is now a string, instead of real

GalacticChimp:
ready for CR and merge


objects/obj_main_menu/Create_0.gml
```diff
@@ -1,9 +1,7 @@
 
 if (instance_number(obj_cuicons)=0){global.custom_icons=4;instance_create(0,0,obj_cuicons);}
 
-
-global.version=0.8000;
-
+global.version="0.8.0.1";
```
GalacticChimp, line 4: might as well go 0.8.0.1, for the first hotfox release

objects/obj_main_menu/Other_2.gml
```diff
@@ -5,6 +5,4 @@ global.icon_name="";
 
 instance_create(0,0,obj_img);
 
-randomize();
-
-script_execute(scr_unit_stats,0,0,0,0,0);
```
GalacticChimp, line 10: this script doesn't do anything at all

objects/obj_main_menu/Step_0.gml
```diff
@@ -33,19 +33,12 @@ if (stage=2) and (tim1<=0){tim2-=1;
     }
 }
 
-
-// Some kind of integrity test through version? global.version set in Create
-if (global.version<version){
```
GalacticChimp, line 38: any integrity tests should be handled during loading saved games. Added a TODO in scr_load, so we won't forget to do that

ZAB909 (approved):

PR #95, opened 2023-09-13
stopped crash not fixed function (bug)

OH296:
- fixed the crash on trying to open the formation changer not fixed the broken functionality though, am going to bed so put here in case anyone get's to it before me


GalacticChimp:
fixed by #85

GalacticChimp:
I'll close this one, since other PRs seem to have fixed the issue. But thanks for taking a shot at this one.

PR #97, opened 2023-09-15
Hotfix/formation bars fixes

felipeStouvenel:
# Fix:
- Fixed drag and drop for formation bars
- Updated and improved code logic

## TODO
- [x] Use unit_type value to update the bar formation
- [x] Fix the click multiple times to select the unit instead of reading the unit value based on mouse hover 
- [ ] The bars are drawn one on top of each other, this should be fixed to draw them next to each other
image


ZAB909:
closing due to #98

PR #109, opened 2023-09-17
moving unit structs

OH296:
- add to logic for moving promoting and transferring marines across companies to support structs


GalacticChimp (changes requested):
looks fine apart from save not working

GalacticChimp (approved):

PR #94, opened 2023-09-13
load button for vets and scouts

OH296:
- adds two buttons on chapter creation allowing scouts and vets to be scattered throughout chapter on game start

PR #110, opened 2023-09-18
save load options bug

OH296:
fixes bug with saving after #94 merge


GalacticChimp (changes requested):
save seems to work, but load crashes (fresh smurf start):

image

GalacticChimp (approved):

PR #113, opened 2023-09-18
fix for inquisitor "witnessing sorcery"

Zorelnam:
Now an inquisitorial ship will need to be over the planet in order to "witness" you using heretical magic

Also, please test, i run my own tests and it looks like it works. Cheers


GalacticChimp:
how to test this? use sorcery without an Inquisitor present  and the event shouldn't popup, and then use sorcery with an Inquisitor present and the event should popup?

Zorelnam:
> how to test this? use sorcery without an Inquisitor present and the event shouldn't popup, and then use sorcery with an Inquisitor present and the event should popup?

In order to test it make a custom chapter or select a pre-made one with "Daemon Binders"
Immediately as soon as you start the game start land battles and attack stuff with your psykers, you should see the normal battles play out and the inquisition won't be alerted
Then wait for an inspection/inquisition ship to move into your system with you and attack stuff again, now you should see the "inquisitor has witnessed using sorcery"

PR #116, opened 2023-09-19
Fix for "Inquisitor Witnessed you use sorcery"

Zorelnam:
In order to test it make a custom chapter or select a pre-made one with "Daemon Binders"
Immediately as soon as you start the game, start land battles and attack stuff with your psykers, you should see the normal battles play out and the inquisition won't be alerted
Then wait for an inspection/inquisition ship to move into your system with you and attack stuff again with your psykers using sorcery, now you should see the message "inquisitor has witnessed using sorcery" and the inquisition will probably call you on the diplomacy screen to complain

For further clarification: Previously the hidden in game counter for "inquisitor witnessed you using sorcery" could increase at any time you use Heretical sorcery, i added one more check to the line, now the game must verify if there is an inquisition ship in orbit before it increases the number.
In other words, in previous (bugged) behaviour the inquisition would always know when you used sorcery even if there was no way they could know.
Now this change should make playing with warp powers viable as long as you avoid using them with inquisition ships in orbit

Edit: woopsie, i re-organized my repository and accidentally deleted the pull request, luckily i had it backed up


GalacticChimp:
this PR has the same code as the previous one?

Zorelnam:
> this PR has the same code as the previous one?

it should be exactly the same, i haven't touched it

GalacticChimp (approved):

PR #121, opened 2023-09-27
Feat: star name refactor

felipeStouvenel:
# Feat
- Re did scr_star_name with improved logic and an array instead of over 500 if statements. This should improve start up performance somewhat

### DIFF
--- a/scripts/scr_star_name/scr_star_name.gml
+++ b/scripts/scr_star_name/scr_star_name.gml
@@ -135,7 +135,7 @@ function scr_star_name() {
 		"Batoria", "Betechton", "Natsigan", "Haboga", "Mard"
 	]
 
-	var ok=0,num="",rand=floor(irandom(array_length(planet_names)-1));
+	var ok=0,num="",rand=irandom(array_length(planet_names)-1);
 
 	for(var i=0; i<100; i++){
 	    if (ok==0){


scripts/scr_star_name/scr_star_name.gml
```diff
@@ -1,595 +1,163 @@
 function scr_star_name() {
+	var rand, ok=0,num="";
+	rand=floor(random(566));
```
GalacticChimp, line 3: instead of `566` I recommend `array_length(planet_names)`, 
also you can just have `var rand = floor(...`, simpler

GalacticChimp (changes requested):
please just resolve my 1 comment and I'll approve and merge

scripts/scr_star_name/scr_star_name.gml
```diff
@@ -1,595 +1,162 @@
...
+		"Woden", "Guderian", "Edelweiss", "Gotenland", "Theodorichshaven",
+		"Batoria", "Betechton", "Natsigan", "Haboga", "Mard"
+	]
 
-	// TODO refactor this to just a list of star names
-	var stib, rand, rand2, ok, num;
-	stib=100;ok=0;num="";
-	rand=floor(random(566))+1;
-	rand2=floor(random(100))+1;
+	var ok=0,num="",rand=floor(irandom(array_length(planet_names)-1));
```
GalacticChimp, line 138: you don't need this `floor`, since we are getting ints from `irandom`

GalacticChimp (changes requested):
just 1 more tiny change

GalacticChimp (approved):

PR #126, opened 2023-10-06
turn marker tool tip

OH296:
- adds a tool tip over the year in the main start map so player can see turn counter, as per dev meeting
- light refactor or scr_ui_popup


OH296:
addressed all points and made a few further QOL changes

OH296:
@felipeStouvenel  ok finished final refactor points

PR #127, opened 2023-10-06
create guidance file

OH296:
adds a basic .txt guidance file for devs to add to.
Basically just the start of a cheat sheet of sorts


code_guidance.txt
GalacticChimp, line 1: looks fine, but I'd suggest .md file format, will allow stuff like checkboxes etc

.DS_Store
GalacticChimp, line 1: also, why are these these included/changed?

OH296: gamemaker binary files, they have no effect on the code base but gamemaker sometimes changes em up after running a test.

code_guidance.txt
OH296: will change

.DS_Store
GalacticChimp: ok, but this is causing a conflict

OH296: interesting...

GalacticChimp: ok, this will fix it: https://github.com/ZAB909/ChapterMaster/pull/131
We don't need .DS_Store files in the repo, but we can keep them locally

GalacticChimp:
@OH296  please merge `main`, it should resolve the conflicts. If not, then let me know, we'll figure it out

PR #117, opened 2023-09-20
Feat: obj_controller refactor 5

felipeStouvenel:
#Feat
- Refactor of obj_controller keyboard_84.gml
- Refactor of obj_controller step_0.gml
- Improved logic
- Added documentation to the scripts

PR #130, opened 2023-10-07
Hotfix: deleted logs which caused slow downs

felipeStouvenel:
# Hotfix
- Deleted a coment which was causing slow downs on system generation

### DIFF
--- a/objects/obj_controller/Alarm_1.gml
+++ b/objects/obj_controller/Alarm_1.gml
@@ -215,9 +215,6 @@ if (did==1){
                 fleet.capital_num[fleet.capital_number]=f;
                 fleet.capital_uid[fleet.capital_number]=obj_ini.ship_uid[fleet.capital_num[fleet.capital_number]];
                 break;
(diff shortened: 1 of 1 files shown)


OH296 (approved):
All looks good.

GalacticChimp (changes requested):
doesn't compile

image

OH296:
> doesn't compile
> 
> image

these are solved by the  #128

GalacticChimp:
> > doesn't compile
> > image
> 
> these are solved by the #128 and were caused by #117 i think

ok, but `ZAB909:main` still needs to be merged/rebased onto this branch

OH296:
> > > doesn't compile
> > > image
> > 
> > 
> > these are solved by the #128 and were caused by #117 i think
> 
> ok, but `ZAB909:main` still needs to be merged/rebased onto this branch
rebased locally and runs fine

GalacticChimp (approved):

PR #176, opened 2023-10-20
unit ui

OH296:
- units now have a full ui viewable by hovering the cursor over the marines sprite
- fix hover over issue with turn counter appearing incorrectly
- hover overs over all the marines stat data with enhanced data,
- marines with tech or librarium potential are highlighted in company ui
- refactored much of initialise_custom (more will com later)
- refactored first company spawn
- made marines spawnable with bionics on game start
- added new traits (started implementing game mechanics with traits for example having the flesh_is_weak trait gives piety stat bonus' when adding a new bionic)
- refactored specialist spawning
- created custom squad specific names for roles (e.g tactical sergeant as opposed to sergeant)
- all marines now have a psionic or pysker value denoting their effect on the warp (if value is high enough can be trained into librarian)
- all marines now have markers for different gene seed mutations, any marine has a chance to spawn with gene seed mutations with the chance increased as a result of chapter creation gene seed stability(makes stability acctually do something)
- fix the CTD crashes from loading and saving (duke was not consistent with starting positions or numbers with arrays so something to look out for)

sorry for the size of this one and having so much different stuff i just kept noticing stuff and got carried away

PR #178, opened 2023-10-23
Wiped out company restock

OH296:
- fixes crash when promoting marines
- fixes marine disappearance when promoting marines
- fixes ui bug where more slots are created than there are units in the company
- fixes promoting bug where unit role does not change
- also contains duplicating vehicle fix

is branch of #176 and will need to be reviewed and merged after it but made separate pr to make prs more concise

### DIFF
--- a/scripts/scr_company_view/scr_company_view.gml
+++ b/scripts/scr_company_view/scr_company_view.gml
@@ -52,9 +52,8 @@ function scr_company_view(company) {
 	sel_uni[1]="Command";
 
 	// This sets up the mans, but not the vehicles
-	var company_length = array_length(obj_ini.name[company]);
 	// Company_lenght is 501, so we check with 500
-	for (var v = 1; v < (company_length-1); v++){
+	for (var v = 0; v <= 500; v++){
 		bad=0;
 
 	    if (company>=0) and (company<=10){
@@ -62,8 +61,12 @@ function scr_company_view(company) {
 	        if (unit.name()!=""){
 				unit_loc = unit.marine_location() ;
 	            // if (obj_ini.god[company,v]>=10) then bad=1;
+				if (unit_loc[0] == location_types.ship){
+				   	if (obj_ini.ship_location[unit_loc[1]]="Lost") then bad=1;
+				}	            
 	            if (bad==1){
 					man[v]="hide";
+					continue;
 				}else{
 	                mans+=1;
 	                man[v]="man";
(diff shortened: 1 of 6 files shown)


scripts/scr_company_view/scr_company_view.gml
```diff
@@ -225,15 +228,15 @@ function scr_company_view(company) {
...
-
-	        if (obj_ini.name[company,v+1]=="") and (obj_ini.name[company,v]!="") and
-	        (last_man==0) and (obj_ini.ship_location[obj_ini.lid[company,v]]!="Lost")
-	        then last_man=v;
+	        if (obj_ini.name[company,v+1]=="")and (last_man==0) and (obj_ini.ship_location[obj_ini.lid[company,v]]!="Lost"){last_man=v;break;}
 	    }
 	}
 
-	v=last_man;
+	v=last_man;sqr(x)
```
GalacticChimp, line 239: no `;`. Also what does this do? just squares an x?

scripts/scr_company_view/scr_company_view.gml
```diff
@@ -52,18 +52,21 @@ function scr_company_view(company) {
 	sel_uni[1]="Command";
 
 	// This sets up the mans, but not the vehicles
-	var company_length = array_length(obj_ini.name[company]);
 	// Company_lenght is 501, so we check with 500
-	for (var v = 1; v < (company_length-1); v++){
+	for (var v = 0; v <= 500; v++){
```
GalacticChimp, line 56: can we keep the old version, i.e. check company length? we really don't want to hardcode stuff anymore

GalacticChimp (changes requested):
looks fine, please just check my comments

scripts/scr_company_view/scr_company_view.gml
```diff
@@ -52,18 +52,23 @@ function scr_company_view(company) {
 	sel_uni[1]="Command";
 
 	// This sets up the mans, but not the vehicles
-	var company_length = array_length(obj_ini.name[company]);
 	// Company_lenght is 501, so we check with 500
-	for (var v = 1; v < (company_length-1); v++){
+	var company_length = array_length(obj_ini.TTRPG[company]);
+	for (var v = 0; v < company_length; v++){
```
GalacticChimp, line 57: shouldn't this start with `v = 1`?

GalacticChimp (changes requested):
1 last question

scripts/scr_company_view/scr_company_view.gml
GalacticChimp: ok, but some lines below, for example, line 76 (right side) you have `ma_race[v]=obj_ini.race[company,v];` on both left and right. Are you sure that changing the starting `v` index won't break those assigments?

GalacticChimp: ok, thanks for the clarification

GalacticChimp (approved):

PR #191, opened 2023-10-30
Refactor weapons and Equipment

OH296:
refactoring weapons to have defined universal stats rather than hardcoding

everything needs converting to string I'Ds but that's gonna need a lot of the main game recoding as the whole game just uses the display names to reference equipment and weapons


GalacticChimp:
this is of course for 0.8.3.0?

JSON_files/gear.JSON
GalacticChimp, line 1: just write json in small letters

.DS_Store
GalacticChimp, line 1: this should not be here, it's ignored in .gitignore?

GalacticChimp: I'll take a look in .gitignore, maybe I typed something wrong

GalacticChimp: this should fix it https://github.com/OH296/ChapterMaster/pull/4

GalacticChimp: open the terminal and run this

`find . -name .DS_Store -print0 | xargs -0 git rm --ignore-unmatch`

should fix it nicely

GalacticChimp:
sure, no need to rush this. For fundamental stuff like this we want to make it good above all else

scripts/scr_weapon/scr_weapon.gml
```diff
@@ -1,85 +1,84 @@
...
-	// argument4: dreaded
-	// argument5: nuum
-	// argument6: description?
+	// equipment_1: name of the first piece of equipment
+	// equipment_2: name of second piece of equipment if any
+	// base_group: the unit type defualts to true which means a marine (this needs work)
+	// unit_array_position: the position of the marine inside the tempory combat array
+	// is_dreadnought: is unit a dreadnought
+	// nuum: not a good god damn fucking clue
+	// information_wanted: what type of information do you want returned
```
GalacticChimp, line 9: you can give bit longer names to args, for example

`equipment_1` -> `first_equipment_piece`
`base_group` -> `unit_base_group` 
etc

IMO in general, if you have to comment vars, then something is wrong

JSON_files/gear.JSON
```diff
@@ -0,0 +1,421 @@
+{
+  "armour": {
+    "Power Armour": {
+      "armour_value": {
+        "standard": 19,
```
GalacticChimp, line 5: these look like quality modifiers, so maybe rename them to `armour_quality_value`? or just `armour_quality`?

JSON_files/gear.JSON
GalacticChimp, line 1: oooh, I really like this one. A single source of truth, for gear and weapons. WAY better than the old implementation.

JSON_files/gear.JSON
```diff
@@ -0,0 +1,421 @@
...
+      "melee_mod": {
+        "standard": 0,
+        "master_crafted": 5, // Augmented
+        "artifact": 10 // Augmented
+      }
+    },
+    "MK10 Indomitus": {
+      "armour_value": {
+        "standard": 24,
+        "master_crafted": 26, // Augmented
```
GalacticChimp, line 246: these comments don't break CM during reading this file? Not all Json serializers support comments

JSON_files/weapons.JSON
```diff
@@ -0,0 +1,2249 @@
+{
+  "Choppa": {
+    "abbreviation": "Chop",
```
GalacticChimp, line 3: good idea to move this here, instead of a separate hardcoded script

JSON_files/weapons.JSON
```diff
@@ -0,0 +1,2249 @@
...
+    },
+    "damage_resistance_mod": {
+        "standard": 0,
+        "master_crafted": 0,
+        "artifact": 0
+    },
+    "description": "More of a sword than a knife proper, this tough and thick blade becomes a deadly weapon in the hand of an Astartes.",
+    "melee_hands": 0,
+    "ranged_hands": 0,
+    "ammo": 0,
```
GalacticChimp, line 128: would be simpler to skip defining props with 0 values and simply set default 0 for these during load

JSON_files/gear.JSON
GalacticChimp: check if removing all comments from .json will fix the issue, the same thing happens in Unity and other apps, unless you use specific json serializers.

objects/obj_main_menu_buttons/Create_0.gml
```diff
@@ -5,7 +5,14 @@ var i;i=-1;
...
     hover[i]=0;glow[i]=0;glowing[i]=1;
 }
-
+var path = working_directory+ "JSON_files/gear.JSON"
+show_debug_message("{0}", path);
+var gear =get_open_filename("*.JSON;", "gear.JSON");
+
+show_debug_message("{0}", gear);
+gear = file_text_read_string(gear);
+global.gear = json_decode(gear);
```
AtlaStar, line 14: Just for reference, `json_encode` and `json_decode` are both deprecated, as they return json data in the form of `ds_maps` and `ds_lists` rather than structs. The modern way is to use `json_parse` and `json_stringify` which returns the data in the form of structs and arrays, and then to set the static struct object if the serialized thing has functions that are statically declared.

Here are the docs for reference https://manual.yoyogames.com/#t=GameMaker_Language%2FGML_Reference%2FFile_Handling%2FEncoding_And_Hashing%2Fjson_decode.htm

PR #194, opened 2023-10-31
specialist search refactor

OH296:
- creates specialised methods for searching for different role sets (e.g pykers or tech based roles) rather than hardcoding if statements.
- improves existing is_specialist method
- for 8.3.0

### DIFF
--- a/scripts/is_specialist/is_specialist.gml
+++ b/scripts/is_specialist/is_specialist.gml
@@ -1,27 +1,85 @@
-function is_specialist(argument0) {
+function is_specialist(unit_role, type="standard", include_trainee=false) {
 
-	// argument0: role
+	// unit_role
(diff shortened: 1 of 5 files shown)


scripts/is_specialist/is_specialist.gml
```diff
@@ -1,27 +1,85 @@
-function is_specialist(argument0) {
+function is_specialist(unit_role, type="standard", include_trainee=false) {
 
-	// argument0: role
+	// unit_role
+	//TODO need to make all string roles not strings but array references
+	var specialist=false;
```
GalacticChimp, line 5:
```suggestion
	var _is_specialist=false;
```

will be a bit clearer imo

GalacticChimp (approved):

PR #198, opened 2023-11-03
solve loading upgrades

OH296:
- fixes loading of planet upgrade
- provides some belt and braces to hopefully solve the one recorded 8.2.1 crash (obj_ini.wid error)


GalacticChimp (approved):
works fine

PR #207, opened 2023-11-08
possible suspect for data leak

OH296:
- done one of two things firstly made sure there are no stray wid variables that could be wiping the 2d array, secondly moved as many places where arrays are referenced like this var[var,var] to var[var][var] as the latter is now preferred by game maker and the former can cause errors.
- this will need vigorous testing to see if there is any practical difference, however they are all good practise changes anyway


GalacticChimp:
there's quite a lot of there [123,123] left. How much do you think we need?

image

GalacticChimp:
yeah, there's 5.6k left, so for now let's fix only those we suspect cause crashes. We can migrate the remaining ones later.

PR #226, opened 2023-11-28
Faction cleanup

AtlaStar:
Replaces areas where a raw value is used as an IDX based on the faction with a new enum called `eFACTION`

Work in progress and needs thorough testing as some stuff was batch replaced, others were manually replaced, and some things may have broken inadvertently


AtlaStar:
> > Also, due to the scope of the changes, I think that it may make the most sense to incrementally change these things and merge them as we go, which is why I stopped working on this branch for now.
> 
> as soon as this is at a point your happy with i'd be happy to make my new branches from here and just keep rebasing or merging as you develop and have a target to merge into main which would simplify things otherwise pushing these changes through is gonna be a pain.

Well I tested some of the things and wasn't running into issues from what I noticed, but it is definitely my opinion that the changes that exist now should be tested and then merged as is, and then have the rest of the changes added incrementally

PR #233, opened 2023-12-05
Promotion black carpace

OH296:
- allows scouts to gain a black carpace during promotion (gives minor stat boost);
- refactor of promotion scripts 
- "promotion" of squads
- full promotion ui refactor


scripts/scr_enemy_ai_a/scr_enemy_ai_a.gml
```diff
@@ -3,10 +3,14 @@ function scr_enemy_ai_a() {
...
+	    var o=0,mx=0,cr=0,tr=0;
+	    repeat(planets){
+	    	o+=1;
+	    	if (p_guardsmen[o]>0) and (cr=0) then cr=o;
+	    }
 	    o=0;
-	    repeat(4){o+=1;
+	    repeat(planets){
+	    	o+=1;
 	        if (p_orks[o]+p_tau[o]+p_chaos[o]+p_traitors[o]+p_tyranids[o]+p_necrons[o]>mx){
```
AtlaStar, line 14: While here, might make sense to include `p_eldar` in the checks or even wrap the logic into a function that gives you the average threat level of non-imperium aligned forces to test against.

scripts/scr_marine_struct/scr_marine_struct.gml
```diff
@@ -614,6 +614,22 @@ function TTRPG_stats(faction, comp, mar, class = "marine") constructor{
 		return obj_ini.role[company][marine_number];
 	};
 	static update_role = function(new_role){
+		if (base_group=="astartes"){
+			if (role() == obj_ini.role[100][12] && new_role!=obj_ini.role[100][12]){
+		  		var has_carpace =false;
```
GalacticChimp, line 619: car**a**pace

scripts/scr_marine_struct/scr_marine_struct.gml
```diff
@@ -614,6 +614,22 @@ function TTRPG_stats(faction, comp, mar, class = "marine") constructor{
...
+		  		if (struct_exists(body[$ "torso"], "black_carpace")){
```
GalacticChimp, line 620: wouldn't it be better to have a separate `add_gene_seed_organ(...)`, then on init/update_role marines would get the organs along with any effects they grant, while other races wouldn't?

scripts/scr_company_view/scr_company_view.gml
```diff
@@ -186,18 +188,17 @@ function scr_company_view(company) {
...
-	                    else if (company==8) and (ma_exp[v]>=60) then ma_promote[v]=1;
-	                    else if (company==7) and (ma_exp[v]>=70) then ma_promote[v]=1;
-	                    else if (company==6) and (ma_exp[v]>=80) then ma_promote[v]=1;
-	                    else if (company==5) and (ma_exp[v]>=100) then ma_promote[v]=1;
-	                    else if (company==4) and (ma_exp[v]>=110) then ma_promote[v]=1;
-	                    else if (company==3) and (ma_exp[v]>=120) then ma_promote[v]=1;
-	                    else if (company==2) and (ma_exp[v]>=150) then ma_promote[v]=1;
-                    
-	                    if (ma_health[v]<=10) then ma_promote[v]=10;
+	                if (is_specialist(unit.role, "rank_and_file")){
```
GalacticChimp, line 191: would be a bit easier and clearer to use, if we had

```suggestion
	                if (unit.role.is_specialist("rank_and_file")){
```
or
```suggestion
	                if (unit.is_specialist("rank_and_file")){
```

scripts/is_specialist/is_specialist.gml
```diff
@@ -142,7 +142,10 @@ function is_specialist(unit_role, type="standard", include_trainee=false) {
...
 		case "trainee":	
 			specialists = role_groups("trainee");
-			break;			
+			break;
+		case "rank_and_file":
+			specialists = role_groups("rank_and_file");
+			break;
 	}
 
 	_is_specialist = (array_contains(specialists,unit_role)) ? true : false;
```
GalacticChimp, line 151: array_contains already returns true/false, so you can just do:

```suggestion
	return array_contains(specialists,unit_role);
```

GalacticChimp:
tried to promote a Salamander Scout into a Devastator into an empty company, but got a crash:

```

___________________________________________
############################################################################################
ERROR in
action number 1
of Mouse Event for Glob Left Button
for object obj_popup:

Variable obj_popup.comp(100763, -2147483648) not set before reading it.
 at gml_Object_obj_popup_Mouse_50 (line 736) -             if (obj_ini.name[target_[comp][i]]=="" and obj_ini.name[target_[comp][i]+1]=="") {
############################################################################################
gml_Object_obj_popup_Mouse_50 (line 736)

```

objects/obj_popup/Mouse_50.gml
```diff
@@ -733,7 +733,7 @@ if (mouse_x>=xx+1465) and (mouse_y>=yy+499) and (mouse_x<xx+1576) and (mouse_y<y
         if (manag>10) then manag=0;
 
         for(i=0;i<501;i++){
-            if (obj_ini.name[target_[comp][i]]=="" and obj_ini.name[target_[comp][i]+1]=="") {
+            if (obj_ini.name[target_comp][i]=="" and obj_ini.name[target_comp][i+1]=="") {
```
GalacticChimp, line 736: I think this change causes the crash

objects/obj_popup/Mouse_50.gml
```diff
@@ -744,188 +744,54 @@ if (mouse_x>=xx+1465) and (mouse_y>=yy+499) and (mouse_x<xx+1576) and (mouse_y<y
...
-                if (req_armour="Power Armour"){
-                    if (array_contains(global.power_armour,obj_controller.ma_armour[i]))  then check=1;
-                }
-                if (req_armour="Terminator Armour"){if (obj_controller.ma_armour[i]="Terminator Armour") or (obj_controller.ma_armour[i]="Tartaros") then check=1;}
-                if (req_armour="Scout Armour"){if (obj_controller.ma_armour[i]="Scout Armour") then check=1;}
-                if (req_armour="Dreadnought"){if (obj_controller.ma_armour[i]="Dreadnought") then check=1;}
 
-                if (check=0){
-                    var satisfied = 0;
+                if (!do_not_change){
```
GalacticChimp, line 749: I recommend not to use negative bools

```suggestion
                if (allow_change){
```

GalacticChimp (changes requested):
some questions and issues to resolve

scripts/scr_company_view/scr_company_view.gml
GalacticChimp: ok, then leave this for another PR

objects/obj_popup/Mouse_50.gml
GalacticChimp: ok, I probably didn't click Clean in GM after switching branches, it works now, sorry for the confusion, however, I spotted another bug, you can promote scouts to scouts, who then lose weapons:

image

scripts/scr_marine_struct/scr_marine_struct.gml
GalacticChimp: for now, just a separate method(s) to update gene_seed stuff, will be clearer and easier to maintain

scripts/is_specialist/is_specialist.gml
```diff
@@ -142,10 +142,13 @@ function is_specialist(unit_role, type="standard", include_trainee=false) {
...
+			break;
+		case "rank_and_file":
+			specialists = role_groups("rank_and_file");
+			break;
 	}
 
-	_is_specialist = (array_contains(specialists,unit_role)) ? true : false;
+	return array_contains(specialists,unit_role);
 
 	return _is_specialist;
```
GalacticChimp, line 153: you don't need this now

scripts/scr_marine_struct/scr_marine_struct.gml
GalacticChimp: > (snip) it might be better to hold the black carapace on the body struct that way we can make it applicable as a mechanic for putting "powered armour" on orks or enhaneced armour on sisters (with the general idea being the black_carapace is simply the area to hold powered or upgraded armour for a unit depending on it's class) with "Black Caraoace" simply being the display name for marines

hmm, interesting idea, but we need to discuss/brainstorm this one. I see a danger of splitting gene seed code all over different areas, which is not something that we want. I'd go either with just keeping body and geneseed separate OR integrating them 100%. I'll create a thread on Discord

scripts/scr_ui_manage/scr_ui_manage.gml
```diff
@@ -954,7 +954,10 @@ function scr_ui_manage() {
 				y5=yy+831;
 				x6=xx+1436;
 				y6=yy+805;
-				draw_unit_buttons([x5,y6, x6, y5],"Promote");
+				if (sel_promoting==1){
+					draw_unit_buttons([x5,y6, x6, y5],"Promote");
```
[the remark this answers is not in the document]
GalacticChimp: maybe make it greyed out? would be clearer that promotion is possible, just not available to the currently selected marine?

objects/obj_popup/Create_0.gml
```diff
@@ -55,9 +55,109 @@ new_target=0;
 
 if (instance_exists(obj_controller)){obj_controller.cooldown=9999;}
 number=0;
+company_promote_data = [
```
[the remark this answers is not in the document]
GalacticChimp: this could also be split into the Company struct PR?

GalacticChimp (approved):

PR #236, opened 2023-12-07
Company struct

OH296:
start organising companies into temporary structs to reduce processing

### DIFF
--- a/scripts/scr_company_struct/scr_company_struct.gml
+++ b/scripts/scr_company_struct/scr_company_struct.gml
@@ -1,0 +1,219 @@
+
+function scr_company_struct(company) constructor{
+	company = company;
+	static find_company_squads = find_company_squads;
+	company_squads = find_company_squads(company);
+	cur_squad = 0;
+	exit_period=false;
+	unit_rollover=false;
+	rollover_sequence=0;
+	selected_unit=temp[120];
+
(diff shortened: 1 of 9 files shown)


scripts/scr_company_struct/scr_company_struct.gml
```diff
@@ -0,0 +1,226 @@
+
+function scr_company_struct(comp) constructor{
```
GalacticChimp, line 2: maybe just name it `Company`?

GalacticChimp: `Company` works, gml is case-sensitive

AtlaStar: Even ignoring the case sensitive aspect of things, this isn't entirely true as you can absolutely shadow variable names, it just means that you have to be more concise with how you access the instance/struct property...as an example

```gml
function Company(company) constructor {
  self.company = company
}
```
`Company` itself becomes a part of the object/struct (or global struct if declared in a script asset), `company` as a local shadows the struct property `company`, which you can then set by just explicitly using self and appending the variable name. This also means you could even use `Company` inside the function as long as it was appended to `self`, it'd just have weird highlighting which is why it isn't recommended.

PR #237, opened 2023-12-08
Most names are now moddable. Harder to have duplicate names.

### LINKED ISSUE #204: Refactor star names handling
Description:
Currently star names are handled by `scr_star_name`, however there's an issue with custom chapters, which fetch recruitment world names from the list. This sometimes results in duplicates which breaks the game, since stuff is mostly identified by it's name currently.

Proposed fix:
Refactor `scr_star_name` to use a hashset/dictionary/whatever internally, which will allow us to track which names are in use and which not. This will also prevent initialiazing the game with duplicate star names, since the new data structure will throw an error immediately upon receiving a duplicate. 

After that, recheck Chapter creation/initialization code to make sure everything will be fine.

GalacticChimp:
The names for the following entities are now stored in .json files and are loaded during game start:

- Sector
- Stars
- Space marines
- Imperials
- Orks
- Eldar
- Tau
- Chaos
- Imperial Ships
- Ork Ships
- Hulks

image

star names can no longer get duplicated. Even if the names run out, the game will simply star naming stars with a generic scheme: Star 81 etc

Fixes #204


objects/obj_creation/Create_0.gml
```diff
@@ -73,8 +73,9 @@ fleet_type=1;
 strength=5;cooperation=5;
 purity=5;stability=5;
 var i;i=-1;repeat(6){i+=1;adv[i]="";adv_num[i]=0;dis[i]="";dis_num[i]=0;}
-homeworld="Temperate";homeworld_name=scr_star_name();
-recruiting="Death";recruiting_name=scr_star_name();
+var starNameGenerator = new StarNameGenerator();
+homeworld="Temperate";homeworld_name=starNameGenerator.generate_random_name();
+recruiting="Death";recruiting_name=starNameGenerator.generate_random_name();
```
GalacticChimp, line 78: hmm, the generator would be easier to do, if I didn't have to instantiate it, just have a 
`StarNameGenerator.generate_random_name()` calls. Not sure how to implement it in GML.

scripts/NameGenerator/NameGenerator.gml
```diff
@@ -0,0 +1,6 @@
+// Script assets have changed for v2.3.0 see
+// https://help.yoyogames.com/hc/en-us/articles/360005277377 for more information
+function NameGenerator() constructor
+{
+	// TODO, thinking how to make name generation bit more generic
```
GalacticChimp, line 5: 95% of the code is same in StarNameGenerator and SystemNameGenerator, need to think a bit how to make it more generic

objects/obj_creation/Create_0.gml
AtlaStar: This would just be accomplished via making `generate_random_name` a static, issue is that you still have to init the static by calling the function or constructor at least once prior to being able to have that syntax style. I would just make `StarNameGenerator` a non constructor function and define that random generator as a static of the function so that `StarNameGenerator` basically acts as a namespace, and then just invoke it to initialize the statics.

scripts/NameGenerator/NameGenerator.gml
```diff
@@ -0,0 +1,6 @@
+// Script assets have changed for v2.3.0 see
+// https://help.yoyogames.com/hc/en-us/articles/360005277377 for more information
+function NameGenerator() constructor
```
AtlaStar, line 3:
```suggestion
function NameGenerator(names_array) constructor
```

alternate is to pass in the string for the path, but this would be useful for making things a bit more generic

scripts/NameGenerator/NameGenerator.gml
AtlaStar:
```suggestion
    static get_random_name = function() {
        return array_pop(names)
    }
    names = array_shuffle(names_array)
    original_names = []
    array_copy(original_names, 0, names, 0)
```
this uses the idea of passing in the loaded JSON rather than a filepath, otherwise the loaded json could just be loaded into `original_names` as a cache and then names could be set to `array_shuffle(original_names)`

objects/obj_creation/Create_0.gml
AtlaStar: Scratch the above, since the goal is to make a generic way to implement the functionality for other areas where names are provided.

scripts/NameGenerator/NameGenerator.gml
GalacticChimp: did something bit similar in `JsonFileListLoader()`, I basically extracted file loading and parsing onto a separate struct, so that the name generators can simply supply a path a list of properties to list

GalacticChimp:
looks like some sprite files got modified as well, I'll revert that

datafiles/main/names/space_marine.json
```diff
@@ -0,0 +1,2078 @@
...
+        "Peetu",
+        "Konsta",
+        "Miro",
+        "Oskari",
+        "Kasper",
+        "Luka",
+        "Juho",
+        "Lauri",
+        "Tapio",
+		"Nelson",
```
GalacticChimp, line 2070: hehe

objects/obj_creation/Create_0.gml
GalacticChimp: ok, I settled with a global `name_generator` struct, which exposes various `generate_imperial_name()` etc. functions.

scripts/NameGenerator/NameGenerator.gml
GalacticChimp: found a way, basically I split name loading from jsons and name composition into different functions.

scripts/NameGenerator/NameGenerator.gml
GalacticChimp: N/A anymore, I chose a different approach

scripts/NameGenerator/NameGenerator.gml
GalacticChimp, line 1: the main struct responsible for most things

scripts/JsonFileListLoader/JsonFileListLoader.gml
GalacticChimp, line 1: I used `file_buffer` instead of `file_text_open_read`, because the former is way faster, and only marginally more complex to use

objects/obj_star_select/Mouse_50.gml
```diff
@@ -190,7 +190,7 @@ if (obj_controller.selecting_planet>0){// Lose focus on no button click
 
 
 if (obj_controller.menu=0) and (obj_controller.zoomed=0) and (!instance_exists(obj_bomb_select)) and (!instance_exists(obj_drop_select)) and (obj_controller.cooldown<=0){
-    var closes,sta1,sta2;closes=0;;sta1=0;sta2=0;
+    var closes,sta1,sta2;closes=0;sta1=0;sta2=0;
```
GalacticChimp, line 193: I also removed a few extra `;` here and there

scripts/scr_add_man/scr_add_man.gml
```diff
@@ -174,11 +173,11 @@ function scr_add_man(man_role, target_company, choice_armour, choice_weapons, ch
...
 	    obj_ini.age[target_company][good]=((obj_controller.millenium*1000)+obj_controller.year);// Age here
     
-	    if (spawn_name="") or (spawn_name="imperial") then obj_ini.name[target_company][good]=scr_marine_name();
+	    if (spawn_name="") or (spawn_name="imperial") then obj_ini.name[target_company][good]=global.name_generator.generate_space_marine_name();
 	    if (spawn_name!="") and (spawn_name!="imperial") then obj_ini.name[target_company][good]=spawn_name;
-	    if (man_role="Ranger") then obj_ini.name[target_company][good]=scr_eldar_name(2);
-	    if (man_role="Ork Sniper") or (man_role="Flash Git") then obj_ini.name[target_company][good]=scr_ork_name();
-	    if (man_role="Sister of Battle") or (man_role="Sister Hospitaler") then obj_ini.name[target_company][good]=scr_imperial_name(2);
+	    if (man_role="Ranger") then obj_ini.name[target_company][good]=global.name_generator.generate_eldar_name(2);
+	    if (man_role="Ork Sniper") or (man_role="Flash Git") then obj_ini.name[target_company][good]=global.name_generator.generate_ork_name();
```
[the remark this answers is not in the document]
GalacticChimp: good idea, but let's do it as a part of another PR

scripts/NameGenerator/NameGenerator.gml
```diff
@@ -0,0 +1,188 @@
...
+	static tau_name_composites = LoadCompositeNames("tau", ["prefixes","suffixes"]);
+
+    // init
+
+    static SimpleNameGeneration = function(names, used_names, entity_name) {
+        if (array_length(names) == 0) {
+            // reset arrays if all names are used up
+            debugl($"Used up all {entity_name} names, resetting name lists");
+            var used_names_length = array_length(used_names);
+            array_copy(names, 0, used_names, 0, used_names_length);
```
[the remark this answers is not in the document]
GalacticChimp: repeats are filtered out in `JsonFileListLoader`:

```
var filtered_data_array = array_unique(property_value);
```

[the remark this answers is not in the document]
GalacticChimp: also, if you see Star Blue 1 it means the names didn't load, so you should check `star.json` validity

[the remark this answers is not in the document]
GalacticChimp: ok, I see what you meant now. Yes, correct, the star names list will get refreshed, I'll add a special check to secure this

scripts/NameGenerator/NameGenerator.gml
```diff
@@ -0,0 +1,188 @@
+function NameGenerator() constructor {
+    static LoadSimpleNames = function(file_name, fallback_value, json_names_property_name = "names") {
```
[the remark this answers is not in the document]
GalacticChimp: in general I agree, but imo it's a bit safer to leave these as static, at least for now. If they are marked as static then we are guaranteed they will be loaded only once. After finishing the save rework we can take a second look at these.

scripts/NameGenerator/NameGenerator.gml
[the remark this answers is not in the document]
GalacticChimp: fixed

image

scripts/NameGenerator/NameGenerator.gml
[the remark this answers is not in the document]
GalacticChimp: added TODO

PR #247, opened 2023-12-28
crusade bug fix, and improvements

OH296:
a few minor bug fixes
- in obj_en_fleet/Alarm1  a variable not being checked for noone value resulting in crash
- necron tombs not displaying the correct name after being sealed or awakening
- excess number of marines sometimes dying in crusades
- also a general refactor of crusades script
- refactored crusade event into a function so it can be activated more readily and cleans up event script
- minor feature in a report of heroic deeds from crusades (logs new traits added to marines)
- added cheat/debug for crusading

to test and to launch crusades at will enter "eventcrusade" into the cheat console and then check the event log to see where the crusade event launched send fleet and end turn until crusade fleet returns

### DIFF
--- a/scripts/scr_crusade/scr_crusade.gml
+++ b/scripts/scr_crusade/scr_crusade.gml
@@ -172,7 +172,7 @@ function launch_crusade(){
 			for(var j = 1; j <= 4 && !assigned_crusade;  j++){
 				if(star_id.p_problem[i][j] == ""){
 					star_id.p_problem[i][j] = "great_crusade";
-					star_id.p_timer[i][j] = 24;
+					star_id.p_timer[i][j] = 36;
 					assigned_crusade = true;
 					break;
 				}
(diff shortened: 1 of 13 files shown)


scripts/scr_crusade/scr_crusade.gml
```diff
@@ -149,15 +139,57 @@ function scr_crusade() {
...
+		debugl("RE: Crusade, couldn't find a star for the crusade");
+		return false;
+	}
+	else{
+		var assigned_crusade = false;
+		for(var i = 1; i <= star_id.planets && !assigned_crusade;i++){
+			for(var j = 1; j <= 4 && !assigned_crusade;  j++){
+				if(star_id.p_problem[i][j] == ""){
+					star_id.p_problem[i][j] = "great_crusade";
+					star_id.p_timer[i][j] = 24;
```
ZAB909, line 175:
```suggestion
					star_id.p_timer[i][j] = 36+;
```
24 is a bit too short.

ZAB909: + of course means go higher if you want

GalacticChimp (approved):

PR #262, opened 2024-01-12
gear not showing in ui equip area

OH296:
solves ui bug with correct gear not showing


jhillacre (approved):
tested, works for me. 👍

GalacticChimp (approved):

PR #307, opened 2024-05-25
Update Step_0.gml to fix mass equip armour

NexiraElimus:
mk 8 is not considered armour, and you cant mass equip individual armours. This fixes both issues. It does not fix there being no option to equip purely terminator armour.


NexiraElimus:
whoops didnt realize pushing changes to my main fork would alter the PR here. Closing this and making a new PR so the mass equip fix can go through on its own while i work on the vehicle equipment/unlocked reward fix.

PR #340, opened 2024-12-27
fix: Stop bug where Squad View no longer appears

OH296:
- attempts to fix the common bug of the squad view becoming inaccessible 
  - it appears it was due to a clause making it non viewable if instances of obj_temp3 existed
- attempt to remove as many uneeded obj_temp3 references as possible that can be done safely and easily

## Summary by Sourcery

Bug Fixes:
- Fix a bug that prevented the Squad View from appearing when instances of obj_temp3 existed.


EttyKitty:
> won't auto merge as i'd prefer a pair of human eyes to look through maybe, tested and runs up and plays fine though so probably all good

Visually, it looks okay. I'll compile and run it for a bit, but that squad view bug is something that I was never able to force spawn.

EttyKitty (approved):
I've tested this for a bit, and it seems to work just fine.

PR #342, opened 2024-12-28
feat: Complex Components for Tartaros Armour

OH296:
- Tartaros right and left leg components
- Tartaros chest components
- new complex area key "gorget"
- Tartaros gorget components

I'm no artist this is as good as i can do i'm afraid


EttyKitty:
I decided against just pushing into this PR, as you may not like what I did, so please check this PR for your PR.
https://github.com/OH296/ChapterMaster/pull/37

PR #356, opened 2024-12-29
feat(sprite): Mk8 components

OH296:
- add gorget component to mk8 and make some variants
- remove gorget from base mk8 so that more mk7 helms are compatible to be drawn with mk8
- add mk8 chest components that can extend variation with mk7 chest components

## Summary by Sourcery

New Features:
- Added new gorget and chest pieces for the Mk 8 armour set.


EttyKitty:
@OH296, I've rechecked some MK8 art, and can I propose using mk7 complex stuff for everything mk8 but gorgets?
By the looks of it, right now everything on mk8 is the same as mk7, and most art is like his as well.

PR #366, opened 2025-01-01
chore: json file error

VanWeapon:
## Description of changes

- adds a popup error if a malform json file is attempted to be read but can't, e.g. someone edits their chapter save and its no longer valid json. 
## Reasons for changes

- Provides instant feedback to players that they made a mistake, can be screenshotted for discord bug reports where a tech priest can inform them how to fix 
## Related links

- https://discord.com/channels/714022226810372107/714027491412410369/1323843628862537802

## Summary by Sourcery

Bug Fixes:
- Detect and report malformed JSON files encountered when loading chapter save data.


VanWeapon:
> Seems to me like for some reason this has fucked up the diplomacy icon images, their numbers should all relate to their faction enum id's but instead i see the ork icon is 13 when it should be 7 and the inquis is instead 7 amongst other issues

@OH296 From testing  #363 it seems to work? I didn't reorder the icon indexes, if you look at `diplomacy_icons.png` which is what i split from, and count boxes from the left starting at 1, then you get the same ids. There's a light and dark version of each small icon which is why the indexes are off

PR #412, opened 2025-01-12
feat: Rework how chapter composition is specified

EttyKitty:
## Description of changes
- Add exp bonus to all marines when the chapter has the Sieged trait (this is how it started).
- Combine all reserve companies to fill the battle ones with the Sieged trait.
- I'm not sure if this counts as a refactor or a feature, but the idea is:
  - Specify all vehicle spawning to not be hardcoded and depend on some variables,
  - Specify unit role **proportions**, instead of hard numbers (for robustness),
  - And probably a bunch of other chapter assembling related shit, will update as needed.
## Reasons for changes
- For the exp buff, it makes some sense that all Sieged chapters are veterans. Perhaps makes more sense to remove the EXP bonus, as they already get a lot of it from the shift into battle companies.
- Company merge makes sense, as I assume this is what happens when the company is reduced below combat efficiency numbers, and it's better to have full Battle Companies.
- Chapter assembling and loops are hectic. In some places it's just a weird hardcoded repeat(4), in some it's a weird counting of how many units of each role is specified and subtracting them from the total unit count. It's just weird.
## Notes
- Perhaps it's better to specify unit spawning in squads (with the default squad sizes) and not individuals, where possible.
- [x] Compile
- [x] New game
- [ ] Next turn
- [ ] Space Travel
- [ ] Ground Battle

### DIFF
--- a/scripts/scr_initialize_custom/scr_initialize_custom.gml
+++ b/scripts/scr_initialize_custom/scr_initialize_custom.gml
@@ -554,26 +554,6 @@ function scr_initialize_custom() {
 
 	// Initializes all of the marine/vehicle/ship variables for the chapter.
 
-	techmarines = 8;
-	apothecary = 8;
-	epistolary = 2;
-	codiciery = 2;
-	lexicanum = 4;
-	terminator = 40;
-	veteran = 70;
-	second = 100;
-	third = 100;
-	fourth = 100;
-	fifth = 100;
-	sixth = 100;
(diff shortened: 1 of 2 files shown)


Blogaugis (commented):
Would it be possible to have Json set the variables?

EttyKitty:
#725 does this. What it doesn't - can be done from the scratch, not worth keeping this here.

PR #450, opened 2025-01-21
 fix: Attempted fix for crash on image id 0

VanWeapon:
## Description of changes
- unsure if this will work for sure, looks like array_create on line 557 was being passed-by-reference and then garbage collected
## Reasons for changes
- attempted crash fix
## Related links
- https://discord.com/channels/714022226810372107/1328975308140974183
- [x] Compile
- [ ] New game
- [ ] Next turn
- [ ] Space Travel
- [ ] Ground Battle


VanWeapon:
The cache array should always exists by the time it is accessed. It is first explicitly checked for, then created if it does not already exist.

PR #471, opened 2025-02-07
fix: Corruption for new recruits

frost923:
## Description of changes
- This update resolves the issue where the new scout's base corruption value was incorrectly set to 10 instead of 0. The corruption is now properly initialized at 0 and hopefully can be reduced as intended.
## Reasons for changes
- The base 10 does not make sense since it's for all types of recruits and cannot be reduced.
## Related links
-
- [X] Compile
- [X] New game
- [X] Next turn
- [X] Space Travel
- [X] Ground Battle

### DIFF
--- a/scripts/scr_recruit_data/scr_recruit_data.gml
+++ b/scripts/scr_recruit_data/scr_recruit_data.gml
@@ -79,7 +79,7 @@ function planet_training_sequence(local_apothecary_points){
 
 	        var recruit_chance = 999;
 	        var aspirant = 0;
-	        var new_recruit_corruption = 10;
+	        var new_recruit_corruption = 0;
 	        var months_to_neo = 72;
 	        var dista = 0;
 	        var onceh = 0;


MCPO-Spartan-117:
I think this might've been intentional given the "apprenticeship" Aspirant Trial is supposed to reduce corruption by `0` to `-10`.

frost923:
Except for some reason it applies to all types of recruitment.

scripts/scr_recruit_data/scr_recruit_data.gml
```diff
@@ -79,7 +79,7 @@ function planet_training_sequence(local_apothecary_points){
 
 	        var recruit_chance = 999;
 	        var aspirant = 0;
-	        var new_recruit_corruption = 10;
+	        var new_recruit_corruption = irandom(10);
```
Blogaugis, line 82: Does the `irandom(n)` function also make it with fractional numbers like 0.005?
This may not be that big of an issue if the game tolerates fractional amounts, but we might want to ensure that they are full numbers rather than fractional.

MCPO-Spartan-117: `irandom` specifically makes integers, not floats.

Blogaugis: Ah, ok. Then you can mark this conversation as 'resolved'.

PR #472, opened 2025-02-07
fix: Rhino missing HK missile armament

frost923:
## Description of changes
- Adds the missing option to add HK missiles after being initialized.
## Reasons for changes
- It was only initialized before but you could not arm them.
## Related links
-
- [X] Compile
- [X] New game
- [X] Next turn
- [X] Space Travel
- [X] Ground Battle


Blogaugis (approved):
Well, I don't see any issues at a quick glance, so here's my approval.
At some point we might want to make it possible for other vehicles too, which can use this weapon.

PR #509, opened 2025-02-24
fix: hand flamer

Blogaugis:
#### Purpose of the PR
Hand flamer had 1.1 range value, which made it a melee weapon. This PR fixes it to have the same range value as infernus pistol.

#### Describe the solution
Increase a number of a value.

Removing hand flamer entirely.
Increasing range values of other pistol weapons.

#### Testing done
- [ ] Compile;
- [ ] Purchase in armamentarium;
- [ ] Combat.

#### Related links
https://discord.com/channels/714022226810372107/1342886304609996870

### DIFF
--- a/scripts/scr_weapon/scr_weapon.gml
+++ b/scripts/scr_weapon/scr_weapon.gml
@@ -1367,10 +1367,11 @@ global.weapons = {
 		"melee_hands": 0,
 		"ranged_hands": 1,
 		"ammo": 2,
-		"range": 1.1,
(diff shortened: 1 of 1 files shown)


Blogaugis:
Test results - I think it needs higher `spli` value - to allow it to damage at least 3, but preferably 5 or 10 enemies with single use.
It should be able to serve as a crowd control weapon.
But otherwise the weapon is functional and could have a niche if more buffs are provided.

PR #516, opened 2025-02-25
feat: ship-stats

Blogaugis:
#### Purpose of the PR
This PR aims to achieve several goals:
Create a low-budged or renegade option for ship building;
Make space combat more viable, especially for renegade chapters.

#### Describe the solution
Introduce new variants of ships for several factions, including the player's.

Make existing ships build-able via alternative means (Such as chapter forge).

#### Testing done
- [ ] Compilation;
- [ ] Ship purchase or forging;
- [ ] Land force loading and unloading;
- [ ] Space travel;
- [ ] Space Combat;
- [ ] Bombardment;
- [ ] Land attack;
- [ ] Artifact and STC collecting.

#### Related links
https://discord.com/channels/714022226810372107/1342830023350157312

### DIFF
--- a/objects/obj_al_ship/Alarm_0.gml
+++ b/objects/obj_al_ship/Alarm_0.gml
@@ -729,6 +729,28 @@ if (class=="Iconoclast"){
     weapon_dam[1]=8;
     weapon_range[1]=300;
     weapon_cooldown[1]=30;
+}else 
+if (class=="Q-ship"){
+    sprite_index=spr_ship_stri;
+    ship_size=2;
+    name="";
+    hp=300;
+    maxhp=300;
+    conditions="";
+    shields=100;
+    maxshields=100;
+    leadership=80;
+    armour_front=3;
+    armour_other=2;
+    weapons=1;
+    turrets=2;
+    capacity=150;
+    carrying=0;
+    weapon[1]="Weapons Battery";
+    weapon_facing[1]="most";
+    weapon_dam[1]=8;
+    weapon_range[1]=300;
+    weapon_cooldown[1]=30;
 }
 // Tyranids
 if (class=="Leviathan"){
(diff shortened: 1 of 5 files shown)


Blogaugis:
> Most important is to fill in the new_player_ship function thats where most of the player ship data is set

Noted, though, I'd like to add these ships to other factions - other imperials, traitors and CSMs - as well.

> So i've been thinking about my preferred method for implementation and i'm thinking of addin a raiding/piracy option to fleets in system which will naturally have the down side of teh imperial navy will zero in on you, we can then use the forge to retrofit captured ships from piracy? Don't know how this sounds to you. I've almost finished my major overhaul of sprites and modding and then i'm thinking i'll make a priority out of this?

Ships are valuable commodity, and I don't really like to abstract them. Still, some abstraction is needed - It would be nice to be able to track factions `transport capacity`, which shows how many merchant ships a faction has, and in turn means the potential the player can earn from raiding/piracy. Not sure if this value should be shown to the player, or visible only in debug mode.
This probably begs for yet another idea discussion.

If we double down on abstraction and simple approach, I imagine this - make use of battle barge trade values (the more valuable the planets, the more potential earnings in requisition per turn) by making piracy earn, let's say, 10 times the BBT, but this nukes the influence, disposition and skyrockets corruption. The requisition acquired can then be used to build ships.


I think we need to make smaller and/or simpler vessels build-able on hive worlds, temperate and worlds that have `shipyard` planetary feature. The violent solutions to earning ships are good, but I think options to build them should be a priority.

Blogaugis:
But frankly, the whole naval code is asking for a refactor of some kind, stuff scattered across numerous files is complicated to keep track of.

Blogaugis:
Backtracking a bit to the simple approach of raiding/piracy:
Player is capable to issue the piracy/raiding order; We naturally have to consider - when the player can issue the order?
Should the player be able to issue the order, when it should logically have no effect (let's say in a system with a single several dead worlds only)? Should the player be able to issue the order on non-imperial controlled worlds?

Then we go to effects: So, supposedly, the imperial fleet will go after the player, if the player engages in piracy/raiding.
How much raiding is needed for navy to get off the Nids that they are likely going to be stuck on? Are we also going to see jank like player raiding ork systems, to see imperials going after them?

Even in simplistic terms, this is a pretty complex issue, and I'm not sure how you plan to make this, to not come off as jank. I'm going to search for a discussion in discord to discuss, or create a new one.

PR #523, opened 2025-02-27
fix: Make spawning new defence fleets work

OH296:
#### Purpose of the PR
Make forge worlds spawning new defence fleets for systems work again, currently they just create dozzens of stacks of fleets that hang over the forge worlds causing memory issues
- add marker in top of screen when in menu view to show haw many active imperial ships (ablative value) and how many the current imperial worlds and forge worlds can support
#### Describe the solution
rewrite from scratch for the most part and abstract to usable functions


EttyKitty:
@coderabbitai review

PR #544, opened 2025-03-04
fix: Disposition change related to fortification

Rathowm:
#### Purpose of changes

Fortifying a planet increases the planet's disposition, but the increase isn't carried into the next turn.
#### Describe the solution

The change wasn't added to planet disposition, but instead to "player_disposition" which was reset every time.

None

#### Testing done

Success in multiple new games.
#### Related links

None

#### Player notes

None


MCPO-Spartan-117:
Looks like to me only line 577 needed the change.

PR #557, opened 2025-03-06
feat: Edit Breacher and Bike squads

EttyKitty:
#### Purpose of changes

- Make them more lore friendly, useful and less of a headache.
- Starting as Lamenters and getting your assault squads replaced with breachers is not fun - they are not proper melee units and just suck.
- Bike squads on the other hand were replacing tactical squads. If we go by the example of the White Scars, the main user of bikes, they have a codex compliant organization, where they maintain 6 battleline squads, 2 close-support and 2 fire-support squads in their battle companies. Bikers cannot be in the battleline squad, but can be in the close-support. Hence, the choice is simple.

#### Describe the solution

- Bikers now take up the assault squad space.
- Breachers now take up the tactical squad space.
- Both have their formation assignments fixed.
- Unnecessary text is removed from the role and squad names.
- Equipment of breachers is adjusted as per the HH TT Breacher Squad reference. They don't use melee weapons.
- Equipment of bikers is adjusted per 7th ed reference, they only use one weapon. I considered giving them pistols or special weapon, but I really don't understand how you're riding a bike and shooting a plasma gun. They can live without it. As they don't have any melee, now they have at least some.

None

#### Testing done

Started a new game as Lamenters, checked the breachers.
Started a new game as a custom chapter with both traits, checked both squads.
All seem to work.

#### Related links

None

#### Player notes

Biker and Breacher squad's equipment is adjusted and their battle roles swapped.

### DIFF
--- a/scripts/scr_initialize_custom/scr_initialize_custom.gml
+++ b/scripts/scr_initialize_custom/scr_initialize_custom.gml
@@ -1997,115 +1997,83 @@ function scr_initialize_custom() {
 	}
 	if (scr_has_adv("Lightning Warriors")) {
 		variable_struct_set(st, "bikers", [
-			[roles.tactical, {
+			[roles.assault, {
 				"max": 9,
 				"min": 4,
 				"loadout": { //tactical marine
 						"required": {
-						"wep1": [wep1[100, 8], 6],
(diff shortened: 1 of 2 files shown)


Blogaugis (approved):
Looks okay to me.

PR #558, opened 2025-03-06
fix: Power Armour tag stripping hirelings

EttyKitty:
#### Purpose of changes

- Sisters and Tech Priests were not getting their PA because of the "power_armour" tag requiring black carapace to use.

#### Describe the solution

- Remove the tag from the hireling armor.

- Adjust the tag check, to include edge cases. Don't want to bother.
- Add a new tag, just for armor that doesn't need the carapace. I don't know about this.
- Add a separate tag "black_carapace" to signify that an armor requires it and lower stats if the unit doesn't have it.

#### Testing done

Started new game, spawned some hirelings with cheats.

#### Related links

https://discord.com/channels/714022226810372107/714023282780799028/1347281774438256790

#### Player notes

Sisters and Tech Priests got their armor back.

### DIFF
--- a/scripts/scr_weapon/scr_weapon.gml
+++ b/scripts/scr_weapon/scr_weapon.gml
@@ -2241,7 +2241,6 @@ global.gear = {
 				"master_crafted": 5,
 				"artifact": 10
 			},
-			"tags": ["power_armour"],
 		},
 	// Sororitas and other imperials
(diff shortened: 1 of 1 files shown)


Blogaugis (approved):
Check for `black_carapace` or `legion` tags seem the best choice in the future.

PR #559, opened 2025-03-06
feat: Remove the slime short-circuit mechanic

EttyKitty:
#### Purpose of changes

- All my friends hate slime.
- The parents of slime hated it as well.

#### Describe the solution

- Destroy the short-circuit mechanic.
- Delete the short-circuit mechanic.
- Remove the short-circuit mechanic.
- Disintegrate the short-circuit mechanic.
- Annihilate the short-circuit mechanic.
- Send the short-circuit mechanic to its grave.

- The mutation still gives considerable disposition debuffs.

Absolutely none. This disgusting shit has to go now.

#### Testing done

None

#### Related links

None

#### Player notes

Slime short-circuit is no more.


MCPO-Spartan-117:
Spot the 10 purity and non-10 stability player.

Blogaugis:
> Describe alternatives you've considered

I think an easy solution is to increase the maintenance cost of equipment carried.

MCPO-Spartan-117:
It does need a auto re-equip, manual is hell.

Blogaugis:
I don't like automatic re-equipping.
How can I be certain that it won't equip the best gear available to player (Mark 4 armour or even artificer), and thus make it prone to being lost?

Blogaugis:
> There is a thing called _role settings_.

You mean this cumbersome option which affects all marines regardless of company the marine is in and whether he has the mutation or not?
Well, yeah, it is a thing. But it is so cumbersome that I have no desire to touch it.
Maybe it could be an option if we introduce role settings for entire company, or even better - based on how mutated/skilled the marine is. But for now, I think you're putting too much hope into this thing.

MCPO-Spartan-117:
> But for now, I think you're putting too much hope into this thing.
Putting to much hope into something that works? wut.

EttyKitty:
> ngl this just seems to be stripping out valid game mechanics. It's anonnoying but as @MCPO-Spartan-117 points out there is a method to not have it. Seems like a major stripping of flavour and uniqueness to remove it

I really don't like the mechanic and there is not a single reference in the lore about the mucranoid mutation producing slime that can destroy power armour.

The influence of this mutation in comparison to other mutations is still too big, even if reequip will be introduced.

And it just sounds silly.
If marines have a mutation that produces slime that destroys armour, they would either mod all of their armour to be resistant to it, so it's a non issue, or not wear armour at all. It's a silly mutation that is just bothersome, very weirs and not flavourful.

Execute all marines with it and call it a day, as they don't just hurt themselves, but your chapter when they destroy pa stockpiles. It's this bad.

Blogaugis (approved):
I agree with Etty here.

EttyKitty:
> nor flavourful? It's literally one of the few bits of flavour that exist beyond a fairly boring simulation.

Fair. Slime is pretty rememberable, if that's what youre talking about.

The problem is that it's always remembered as a stupid thing that happenes, as I haven't seen anyone giving good (opposing the deletion) opinions about this mutation at any point a discussion has started about removing it (may be wrong, but I actually only remember bad things).

If the flavour = funny battle event, then 90% of other mutations fail this check. Just an argument for the fact that there is no hard need for extreme flavourlness in the current state of mutations, as it just overshadows all other ones.

> If one really hates that it can appear randomly make it a mutation that only spawns when that mutation is chosen rather than being included in the random assignments to marines based on stability.

Then no one will be picking this mutation ever.
If people are executing marines to get rid of this, the least they can do is just avoid it when possible.

> as for lore, gonna hard disagree there. The entire point of mutations is they're flavour box randomness. It's generic 40k that geneseeds in 40k are degenerating shit shows the exact effects of these degenerations are prime rpg fodder for a game like this to take liberalities with otherwise i'm sorry this just devolves in to the most boring display of a flavourless simulator by which all outcomes can be completely predicted which is in my eyes just kills it.

The argument was not as much about "this mutation can't happen in 40k", but "destroying power armour from inside by a short-circuit with slime" sounds like too of a big lore stretch.

Mucranoid overproducing oil/slime? Sure. It destroying armour? Around 90% sure is impossible, both for reasons that pa is probably not a tv remote, and that, again, marine chapters with this mutation would adapt either by modifying their armour or by not weiring it at all.

Or they will get slaughtered and destroyed as a chapter, as it's unfeasible to survive while just bearing with it as it is.

> In conclusion is there a playability issue regarding balance : yes. Is bulk removing code that will never be replaced the answer: no.

I agree with only the fact that perhaps just turning the mutation into a disposition devuff is not enough.

It's mentioned in the lore that the mucranoid gives protection from extreme heat, cold and void combat.
So I can just cook up debuffs for these things (planet type and boarding).

EttyKitty:
Closed, because #646 addresses the issue better.

PR #574, opened 2025-03-09
feat: Buff sisters of battle

CassieCaster:
Quick and dirty update to the sisters of battle to bring them more in line with Skitarii.

#### Purpose of changes

Right now the skitarii have 2 ranged and melee burden, while the sisters have one. I think this is because the Skitarii think that they're part of the "skitarii" base group. There is no such group.

#### Describe the solution
I gave the sisters 2 burden. It's quick, it's dirty. It doesn't solve the Skitarii problem.

I tried to do a thing where the calculations for hands would actually be 1+(strength/50), if the sister had less then 50 strength. 
That failed. So I then attempted to do a thing where if she had less then 15 strength she'd have 1.5 burdens instead, that failed, as it defaulted to 2. That also failed, as it defaulted to 2.

#### Testing done
Booted the game up in Game maker. It worked.

#### Related links

None

#### Player notes
Small buffs to the sisters of battle.


Blogaugis (approved):
I think it might be better to buff their worn armor instead, but I guess this'll do.
Stats are a bit much if you ask me, but I guess they are due for a review at some point anyway.

CassieCaster:
The... stats are... a bit much? They're not even even brought in line with the Skitarii. They're just higher then they were. Overall they're still weaker then them, when they should be stronger.

I changed the hands_limits on the human profile to allow them to use combos like shields & melee weapons at the same time. Allowing for tactical flexibility. I _wanted_ to give them 1.9 or 1.5 + some scaling, but my coding skill was lacking.

Blogaugis:
> The... stats are... a bit much? They're not even even brought in line with the Skitarii. They're just higher then they were. Overall they're still weaker then them, when they should be stronger.
> 
> I changed the hands_limits on the human profile to allow them to use combos like shields & melee weapons at the same time. Allowing for tactical flexibility. I _wanted_ to give them 1.9 or 1.5 + some scaling, but my coding skill was lacking.

Skitarii are heavily augmented cyborgs.
Sisters are supposedly mostly un-augmented humans (10, 10, 10, 10 stats for a chad IGuardsman), but maybe I'm wrong, or maybe it is a gameplay consideration to make them more useful as hireling.

Blogaugis:
Oh, and @CassieCaster or @EttyKitty do fix the title, I suggest:
fix: buff sisters of battle
So that PR title is going to be validated and consistent with rest of Pull Requests.

EttyKitty:
ngl, I'm not sure about the hands increase.
I don't remember if that was me who took it down or if it was like that already, but you're not supposed to customize your auxiliary units at all, and even worse with space marine weapons.
2 hands mean that they can now equip normal bolters and basically any loadout that a space marine can.

If the goal is to make them similar to skitarii or tech-priests - I advise brining down hands on these units instead, together with their weapon burden, if needed.

CassieCaster:
I didn't want to increase the hands to 2. I wanted it to be 1.5 to keep bolters and stuff like that out of their hands. Sadly, I the game kept rounding it to 2, so I gave up.

I don't agree with the idea that you're not supposed to customize your auxiliary units. A sister of battle is just as capable of wielding chain swords, power swords and a variety of other equipment as marines.

I think the ideal solution is actually the following. Give Skitarii and Sisters hands of 1.75 and bring down the burden of the shields to 0.75 (from .9). This would allow sisters to use shields with a variety of equipment, while keeping bolters and other large equipment out of their hands.

Additionally I would increase the burden of the light bolter to 1.25, so that they can't use both their light bolter, and a shield as that would bring them above the cap burden. Instead they'd have to pick a one handed weapon.

CassieCaster:
> That is weird, and I don't think is intentional.
I agree, and it may very well be a limitation of my skill to a degree. But it kept happening.
> We even talked about locking aux units from any weapon or armor edits. These are here just to add some flavor and utility, you're not supposed to customize or overuse them.
I was not a part of any such conversation. I don't agree with that viewpoint, as I think that customizing units, _any_ units is fun.
I would rather see that expanded then removed.
>Their columns are also not editable, so you can't even separate the melee units from the ranged.
I know, and I don't mind this. If you've got a weird hodgepodge of auxiliaries with you, I think that them struggling to shine when they're deploying beside space marines is perfectly fine. I think it's fine that the player has to figure how how they want to make the most use out of their hirelings and in what way.
>The alternative is to add small variants of all such weapons, which is a possible bloat issue
Yes. I mean, that is an alternative. But I don't think it's required. A hireling will always be weaker then a space marine, simply because of the vast stat gulf.

I know that there were talks about changing the hirelings into be squads rather then individuals. I still agree with that in spirit, but even if they're squads I think that they should be customizable to a degree.

scripts/scr_marine_struct/scr_marine_struct.gml
```diff
@@ -255,7 +255,7 @@ global.base_stats = { //tempory stats subject to change by anyone that wishes to
 				weapons:{
 					"Light Bolter":1, "Bolt Pistol":1, "Chainsword":1, "Sarissa":1}},
 			start_gear:{"armour":"Light Power Armour", "wep1":"Light Bolter", "wep2":""},
-			base_group : "human",
+			base_group : "",
```
Blogaugis, line 258: Hey, why are SoBs not part of human group anymore?

CassieCaster: What the...

scripts/scr_marine_struct/scr_marine_struct.gml
```diff
@@ -1132,7 +1132,7 @@ function TTRPG_stats(faction, comp, mar, class = "marine", other_spawn_data={})
 			} else if base_group == "tech_priest" {
 				ranged_hands_limit = 1+(technology/100);
 			}else if base_group == "human" {
-				ranged_hands_limit = 1;
+				ranged_hands_limit = 1.75;
```
Blogaugis, line 1135:
```suggestion
				ranged_hands_limit = 1.75+(dexterity/40)+(ballistic_skill/80);
```
As in #576 , I suggest adding a natural, even if weak, buff.

scripts/scr_marine_struct/scr_marine_struct.gml
```diff
@@ -1311,7 +1311,7 @@ function TTRPG_stats(faction, comp, mar, class = "marine", other_spawn_data={})
 			} else if base_group == "tech_priest" {
 				melee_hands_limit = 1+(technology/100);
 			}else if base_group == "human" {
-				melee_hands_limit = 1;
+				melee_hands_limit = 1.75;
```
Blogaugis, line 1314:
```suggestion
				melee_hands_limit = 1.75+(strength/40)+(weapon_skill/80);
```
And this one.

CassieCaster:
Thank you for letting me know that for some reason I managed to remove the sister's humanity. 

That said though, I'm not adding a scaling buff to their hands capacity. As it stands, they can wield everything I want them to wield.

scripts/scr_marine_struct/scr_marine_struct.gml
```diff
@@ -1311,7 +1311,7 @@ function TTRPG_stats(faction, comp, mar, class = "marine", other_spawn_data={})
 			} else if base_group == "tech_priest" {
 				melee_hands_limit = 1+(technology/100);
 			}else if base_group == "human" {
-				melee_hands_limit = 1.75;
+				melee_hands_limit = 1.75+(strength/40)+(weapon_skill/80);
```
EttyKitty, line 1314: does this by any chance take them over or close to 2?
p.s. sob got fluid hand limits before marines. they are trule the protagonists of this game.

EttyKitty:
@coderabbitai review

PR #580, opened 2025-03-10
feat: Custom priority and costs to vehicle recovery

EttyKitty:
#### Purpose of changes

Previously, chances to recover a bike were the same as to recover a land raider. Needless to say, I don't need saving bikes, while sacrificing land raiders.
Costs to recover each type of vehicle was also the same, just 1 point of recovery points.

#### Describe the solution

Vehicles have custom priorities and recovery costs (for now, both are the same value).
```
        var _vehicles_priority = {
            "Land Raider": 10,
            "Predator": 5,
            "Whirlwind": 4,
            "Rhino": 3,
            "Land Speeder": 3,
            "Bike": 1
        }
```
Techmarine recovery points calculation was also changed to account for the new cost system.
```
                skill_level = unit.technology / 10;
                if (marine_mobi[i]=="Servo-arm") {
                    skill_level *= 1.5; 
                } else if (marine_mobi[i]=="Servo-harness") {
                    skill_level *= 2;
                }
                skill_level += random(unit.luck / 2);
                obj_ncombat.vehicle_recovery_score += skill_level;
```

None

#### Testing done

A couple of battles. Points are calculated correctly and vehicles are recovered. A bit hard to test each vehicle.

#### Related links

https://discord.com/channels/714022226810372107/1348311762629558282

#### Player notes

Vehicles now have custom priorities and costs for recovery after being lost.

### DIFF
--- a/objects/obj_pnunit/Alarm_5.gml
+++ b/objects/obj_pnunit/Alarm_5.gml
@@ -53,7 +53,7 @@ if (obj_ncombat.defeat == 0) {
     
             if (obj_controller.stc_bonus[3] = 4) {
                 _survival_roll = 80 + _vehicles_priority[$ veh_type[i]];
-                _dice_roll = roll_dice(1, 100, "high");
+                _dice_roll = irandom(1, 100);
                 if (_dice_roll > _survival_roll) && (veh_dead[i] != 2) {
                     veh_hp[i] = 10;
                     veh_dead[i] = 0;


objects/obj_pnunit/Alarm_5.gml
```diff
@@ -28,26 +28,47 @@ if (obj_ncombat.defeat == 0) {
...
+            "Rhino": 3,
+            "Land Speeder": 3,
+            "Bike": 1
+        }
+    
+        for (var i = 0; i < array_length(veh_dead); i++) {
+            if (struct_exists(_vehicles_priority, veh_type[i])) && (veh_dead[i]) && (!veh_ally[i] ) {
+                ds_priority_add(_dead_vehicles_pq, i, _vehicles_priority[$ veh_type[i]]);
+            }
+        }
```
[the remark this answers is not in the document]
EttyKitty: `variable_struct_exists` checks if the provided key has a value. `struct_exists` if the key exists in the first place.
Unless you're sure that `variable_struct_exists` also checks if the key exist (which I'm not sure about), it'll throw an error if it doesn't, and that's what we're checking.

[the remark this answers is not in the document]
EttyKitty: Well, it actually does exist:
https://manual.gamemaker.io/beta/en/GameMaker_Language/GML_Reference/Variable_Functions/variable_struct_exists.htm

In your analysis, you've used the LTS version of the manual, from what I can see. It's very outdated.
Use either the beta or the monthly one.

objects/obj_pnunit/Alarm_4.gml
```diff
@@ -37,14 +38,16 @@ for (var i=0;i<array_length(unit_struct);i++){
...
+                skill_level = unit.technology / 10;
                 if (marine_mobi[i]=="Servo-arm") {
-                    skill_level*=2; 
+                    skill_level *= 1.5; 
                 } else if (marine_mobi[i]=="Servo-harness") {
-                    skill_level*=4;
+                    skill_level *= 2;
                 }
-                skill_level += random(unit.luck*0.05);
+                skill_level += random(unit.luck / 2);
```
Blogaugis, line 50: I have an idea to add `simple tools` and `combi-tool` options as part of #584 in the future.
I guess `combi-tool` would translate directly, with no modifiers, while simple tools would halve the skill level of the marine.

EttyKitty: > I have an idea to add `simple tools` and `combi-tool` options as part of #584 in the future. I guess `combi-tool` would translate directly, with no modifiers, while `simple tools` would halve the skill level of the marine.

I don't think there is a need for that, as we already assume that, if you mean a gear item. Gear slots (1) are very limiting. Before that's solved, I doubt anyone will bother with these tools.

Blogaugis: It is related to that: https://discord.com/channels/714022226810372107/1349114609696636982

EttyKitty:
@coderabbitai, create a new issue related to this pr, I found a mistake in the code.
The recovery priority code runs for every unit column separately at the end of battle, not for all player forces. So the priority queue doesn't work properly.
The solution is to move the whole priority queue logic to the `obj_ncombat` combat end alarm, but how to send references to each unit from the array, in each column - I'm not sure.

PR #581, opened 2025-03-10
feat: Trait influenced stat growth

OH296:
#### Purpose of changes
make a marines individual traits effect their stat growth as well as make a generally better system for stat growth 
#### Describe the solution

- so currently the formula is this every trait adds that a marine has is scanned through and the scores it has in each stat are tallied.
    - if a trait has a guass score but no modifier e.g no max or min the guass number is ignored .
    - if it does have a modifier the score is also added to the tally. so mins have double negative effects max positives
- once these are tallied the scores of each is divided by 4 than floored
- from the final score the base stat growths of the given unit then get boosts so in a marines case WIS, WS and BS,
- from teh final scores a percentage chance for each stat is then calculated
- also in general refactored code for displaying stats

None

#### Testing done

None

#### Related links

None

#### Player notes

None

### DIFF
--- a/scripts/exp_and_exp_growth/exp_and_exp_growth.gml
+++ b/scripts/exp_and_exp_growth/exp_and_exp_growth.gml
@@ -46,7 +46,7 @@ global.stat_display_colour = {
 
 global.stat_icons = {
 	"constitution":spr_constitution_icon, 
-	"strength":spr_dexterity_icon, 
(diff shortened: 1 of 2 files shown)


Blogaugis:
Seems someone caught my (or had a similar) idea.
Nice.
I'll approve when it's ready.

PR #585, opened 2025-03-12
feat: Merge Purge and Attack Bombard

MCPO-Spartan-117:
### Purpose of changes
Title.

### Describe the solution
By merging.

### Testing done
Load game, bombard, purge, attack.

### Custom player notes entry

Use the PR title.

----

* Kills are heavily nerfed, partly due to capital ships losing a point
* Ships that don't have marines are missing from the list
* Ships that don't give any points are in the list


MCPO-Spartan-117:
> why though?

Because they share 80% of code and having them separate is annoying to deal with more complex scenarios, plus it will fix https://discord.com/channels/714022226810372107/1336762209900171305 and some more unexpected things, https://discord.com/channels/714022226810372107/714023282780799028/1348417288797294644.

MCPO-Spartan-117:
We could do that, though we'd need to think about where to place it, 4th button is taken by torpedoes and the old recruiting thing.

PR #587, opened 2025-03-12
Load to ships by default

EttyKitty:
### Purpose of changes

Load to ships is default when you enter the creation screen. Makes sense imo for it to be default for json as well.
Allows to skip the load to ships code chunk on most chapters.

### Describe the solution

Load to ships as default for json chapters.

None

### Testing done

None

### Related links

None

### Custom player notes entry

Use the PR title.

### DIFF
--- a/scripts/scr_chapter_new/scr_chapter_new.gml
+++ b/scripts/scr_chapter_new/scr_chapter_new.gml
@@ -87,7 +87,7 @@ function ChapterData() constructor {
 	battle_cry = "For the Emperor";
 	equal_specialists = 0;
 	load_to_ships = {
-		escort_load: 0,
+		escort_load: 2,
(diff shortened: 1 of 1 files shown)


EttyKitty:
> > Allows to skip the load to ships code chunk on most chapters.
> 
> what does this mean?

all chapter json files have this chunk atm.
```
"load_to_ships": {
            "escort_load": 2, // 0 no, 2 yes, 1 doesnt do anything :)
            "split_scouts": 0, // 0 no, 1 yes
            "split_vets": 0 // 0 no, 1 yes
        },
```
it's copy-pasted and is exactly the same everywhere. if the PR is merged, the whole thing can be deleted from all json chapters.
just less retyping and line bloat.
the de-facto default is 2. it's only json default has it at 0.

PR #594, opened 2025-03-15
fix: Forge imperium ships merging

MCPO-Spartan-117:
### Purpose of changes
Allow Imperium forge ships to merge.

### Describe the solution
`string_count("!",trade_goods)>0` to `string_count("merge", trade_goods) > 0`

### Testing done
Load game, next turn.

https://discord.com/channels/714022226810372107/1121959429546455050/1350271258855477269

### DIFF
--- a/objects/obj_en_fleet/Collision_obj_en_fleet.gml
+++ b/objects/obj_en_fleet/Collision_obj_en_fleet.gml
@@ -2,15 +2,15 @@ if (last_turn_check == obj_controller.turn){
     exit;
 }
 
-var _same_navy = navy == other.navy;
-if (other.owner == self.owner && _same_navy) {
-    if !(action_x == other.action_x && action_y == other.action_y) { exit; }
-
-    if (trade_goods != "" && other.trade_goods != "") && (!fleet_has_cargo("colonize") && !fleet_has_cargo("colonize", other)) {
-        if (action_x == other.action_x && action_y == other.action_y) && (!fleet_has_cargo("ork_warboss") && !fleet_has_cargo("ork_warboss", other)) { // ork_warboss would never match as it seems only imperium ships get the 'merge' directive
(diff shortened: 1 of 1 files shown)


MCPO-Spartan-117:
@coderabbitai review

MCPO-Spartan-117:
@coderabbitai review

MCPO-Spartan-117:
Bro, there are 0 alien fleets in my save, and it is whatever ships that come out of forge worlds, they spawn with the directive "merge" and they try to go through to object to do so, but it was looking for "!" which is for inquisition ships meant for the player being late.

5132 calls went down to 94.

MCPO-Spartan-117:
Only instance of merge.
`scr_forge_world_functions`
```csharp
        if (_defence_fleet && array_length(_value_hierarchy)){
        	_current_imperial_fleet.trade_goods = "merge";
```

MCPO-Spartan-117:
> travel destination which occurs in fleet_arrival_logic

Which requires the fleets to have the "move" action in both places it is executed, which seems to be only set when the function "set_fleet_movement" with the argument "fastest_route" is set to "false" unless it fails which then sets it, the forge function doesn't provide a argument and it defaults to "true".

MCPO-Spartan-117:
Are or are they not supposed to merge, if not then WHY do they have the merge directive?

There is only one place where they get this directive.

MCPO-Spartan-117:
And now after looking at this commit https://github.com/Adeptus-Dominus/ChapterMaster/commit/7af89cc5d0a91573fbb79407a0d077ed97a9d46f i no longer understand why you even brought this up.

It seems they properly merge after this and in the old save there are still ships doing old behavior, new ships merge with other new ships and move.

So why are you asking this question if they should merge when you are the one that fixed it to begin with?

PR #606, opened 2025-03-19
feat: Adjust how the starting disposition is affected by the sliders

EttyKitty:
### Purpose of changes

- In the original formulas, cooperation slider was not very good imo, both if you scroll it down or up. Free 10 points.
- Disposition was only affected by the purity, when stability also means mutated marines, even worse, actually.
- Strength was only reducing disposition when above 5, not increasing when below. A bit weird.

### Describe the solution

- Each point of cooperation now gives/takes 4 disposition with all factions, except inquisition and mechanicus. They get 2 per point.
- Each point of purity below 10 reduces disposition with inquisition by 2, with ecclesiarchy by 1.
- Each point of stability below 10 reduces disposition with inquisition by 2, with ecclesiarchy by 1.
- Each point of strength below 5 now increases imperium disposition by 2.
- Daemon Binders now also reduces the church dispo by 8.
- Psyker related traits also reduce/add the church disposition, in addition to inquisition.
- There is now an enum to ease readability.

None

### Testing done

None

### Related links

None

### Custom player notes entry

Use the PR title.

### DIFF
--- a/scripts/scr_creation/scr_creation.gml
+++ b/scripts/scr_creation/scr_creation.gml
@@ -113,14 +113,19 @@ function scr_creation(slide_num) {
 				disposition[eSTART_FACTION.Inquisition] = 30 + ((cooperation - 5) * 2) - (2 * (10 - purity)) - (2 * (10 - stability)); // Inq
 				disposition[eSTART_FACTION.Ecclesiarchy] = 40 + ((cooperation - 5) * 4)  - (10 - purity) - ((10 - stability)); // Ecclesiarchy
 			
-				if (founding == eCHAPTERS.SPACE_WOLVES) {
-					disposition[eSTART_FACTION.Progenitor] = 70;
-				} else if (founding == eCHAPTERS.IMPERIAL_FISTS) {
-					disposition[eSTART_FACTION.Progenitor] = 50;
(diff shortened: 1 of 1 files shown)


Blogaugis (approved):

scripts/scr_creation/scr_creation.gml
```diff
@@ -113,14 +113,19 @@ function scr_creation(slide_num) {
...
-				if (founding == eCHAPTERS.SPACE_WOLVES) {
-					disposition[eSTART_FACTION.Progenitor] = 70;
-				} else if (founding == eCHAPTERS.IMPERIAL_FISTS) {
-					disposition[eSTART_FACTION.Progenitor] = 50;
-				} else if (founding == eCHAPTERS.SALAMANDERS) {
-					disposition[eSTART_FACTION.Progenitor] = 70;
-				} else if (founding == 10) { // random/unknown
-					disposition[eSTART_FACTION.Inquisition] -= 5;
+				switch (founding) {
+					case eCHAPTERS.SPACE_WOLVES:
```
Blogaugis, line 117: Are space wolves supposed to be missing stuff here?

Blogaugis (approved):

PR #482, opened 2025-02-14
feat: Strength Rating Changes

Blogaugis:
## Description of changes
- Increases the Strength Ratings of foes and allies from 1-6 to 1-10;
- Changes Strength Rating naming from vague old one to the Space Marine estimated formation needed to defeat them;
- Expanded the requisition reward table to accommodate the increased ranges of Strength Ratings.
## Reasons for changes
- The 1-6 Strength Ratings had a tendency to suddenly become very costly to fight. They also were rather inconsistent in terms of their combat power (orks for example on SR 1 required at least a squad to take them on with a reasonable chance of success, but Tyranids on SR 1 could be defeated by a single marine);
- Changing the names from old variant to Space Marine estimated formation gives player a rough idea of the amount of forces needed to defeat the foes.
## Related links
- https://discord.com/channels/714022226810372107/1309480267165204581
- https://discord.com/channels/714022226810372107/1316042049170182144
## How have you tested your changes?
- [x] Compile
- [x] New game
- [x] Next turn
- [ ] Space Travel
- [x] Ground Battle

PR #607, opened 2025-03-19
Don't repeat "presence" 6 times thanks

EttyKitty:
### Purpose of changes

- Having presence in each line takes too much space and is repetitive.
- Some faction names are undescriptive.

### Describe the solution

- Planet Features > Planetary Features (just for consistency with the latter).
- Planet Forces > Planetary Presence, just because "Planet Presence" sounds weird, I think?
- Ork Presence:  Minima (1) > Orks: Minima (1).
- Adepta Sororitas > Adeptas. It has no reason to be this long and clip into planetary features.
- Traitors > Chaos, as that's what it is.
- CSM > Traitors.
- The code is refactored a bit.
image

None

### Testing done

Checked how it looks in the game.

### Related links

None

### Custom player notes entry

Use the PR title.

### DIFF
--- a/scripts/scr_PlanetData/scr_PlanetData.gml
+++ b/scripts/scr_PlanetData/scr_PlanetData.gml
@@ -627,39 +627,39 @@ function PlanetData(planet, system) constructor{
         
         
         draw_set_font(fnt_40k_14b);
-        draw_text(xx+349,yy+326,"Planet Forces");
+        draw_text(xx+349,yy+326,"Planet Presence");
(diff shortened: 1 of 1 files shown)


Blogaugis (approved):
I guess this simplifies #482 as now all these names are in single place, right?

EttyKitty:
> I guess this simplifies #482 as now all these names are in single place, right?

Yeah, shouldn't clash too much. You can just add new names into the array, at the end, and edit the loop below where I said "if above 6 - use name 6" to up to 10.

PR #615, opened 2025-03-21
fix: Disable enemy array compressing, as it's broken

EttyKitty:
### Purpose of changes

- It doesn't work properly. Enemy arrays get emptied out and deleted in the subsequent empty column destruction, causing enemy columns to shrink after one of the enemy types got killed.
- I tried debugging `dudes_num` and it's for some reason broken as well, showing wrong unit counts. This is the reason for why the array gets cleaned - because it should.

### Describe the solution

- Disable the array compressing for now.
- Refactor the compressing function to just delete empty members, instead of setting them to default. 

None

### Testing done

A battle against orks where I previously observed the bug. No problems anymore.
The array compressing was supposedly only helping with performance, by reducing various "find me something" loop times.

### Related links

None

### Custom player notes entry

Use the PR title.

### DIFF
--- a/scripts/scr_clean/scr_clean.gml
+++ b/scripts/scr_clean/scr_clean.gml
@@ -7,59 +7,22 @@ function compress_enemy_array(_target_column) {
 		return;
 	}
 
-	with (_target_column) {
-		// Define all data arrays to be processed with their default values
-        var _data_arrays = [{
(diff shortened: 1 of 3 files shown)


EttyKitty:
@coderabbitai review

PR #624, opened 2025-03-23
fix: Unit management view breaking and win some step time

EttyKitty:
### Purpose of changes

- Fix a bug that was causing the management view to open with the chapter view on top.
- Reduce step time on the management screen.

### Describe the solution

- Move the scr_management call, so that "apparently" it's not called when not needed.
- Improve step performance by precalculating most of the performance intensive unit view stats once, instead of every step.
- Kill some show_debug stuff, because it was flooding the output.

Kill the `max_health` calls on the management view, as it eats 3ms by just existing for 33 units on the screen every step.

### Testing done

Both fleet and system unit management now opens properly.
Step time on the management screen is improved by ~3-4 ms.

### Related links

https://discord.com/channels/714022226810372107/1353143470923059290

### Custom player notes entry

Use the PR title.

### DIFF
--- a/scripts/scr_draw_management_unit/scr_draw_management_unit.gml
+++ b/scripts/scr_draw_management_unit/scr_draw_management_unit.gml
@@ -4,7 +4,7 @@ function scr_draw_management_unit(selected, yy = 0, xx = 0, draw = true) {
     var string_role = "";
     var health_string = "";
     var eventing = false;
-    jailed = false;
+    var jailed = false;
     var impossible = !is_struct(display_unit[selected]) && !is_array(display_unit[selected]);
     var is_man = false;
     if (man[selected] == "man" && is_struct(display_unit[selected])) {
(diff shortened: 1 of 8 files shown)


scripts/scr_draw_management_unit/scr_draw_management_unit.gml
EttyKitty, line 1: @coderabbitai, can you propose any (preferably quick and simple) performance improvements to this script? It's run on average 33 times per step, to display info about 33 units in the list on the screen, impacting the step time by about 15ms.

EttyKitty: TBF, the main bulk of performance impact comes from the fact that this function is run every step, 33 times.
The info it displays is mostly static, and is updated rarely, so the proper way to address the performance, is to precalculate everything that this function displays and then draw it.
Then call some update function each time any action that may change the displayed info is taken.
Micro-optimizations to the function won't help much, I don't think.

PR #640, opened 2025-03-27
feat: Adjust after battle EXP and Req rewards

EttyKitty:
### Purpose of changes

- To make the reward table a bit more predictable and simple.

### Describe the solution

- Exp and Req reward formulas are now `_quad_factor * sqr(threat);`. Exp has the factor of 10, and Req of 6.
- In practice, this means that the rewards are a bit smaller at lower levels, but higher at high levels.
image

None

### Testing done

None

### Related links

None

### Custom player notes entry

Use the PR title.

### DIFF
--- a/objects/obj_ncombat/Alarm_5.gml
+++ b/objects/obj_ncombat/Alarm_5.gml
@@ -442,7 +442,7 @@ if (defeat=0) and (reduce_power=true){
         new_power = max(new_power, 0);
 
         // Give some money for killing enemies?
-        var _quad_factor = 8;
+        var _quad_factor = 6;
         requisition_reward = _quad_factor * sqr(threat);
         obj_controller.requisition += requisition_reward;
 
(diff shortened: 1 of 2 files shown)


EttyKitty:
Actually, can probably get away with reducing the factor to 4-6.

Blogaugis:
Have to note of potential conflicts with #482

objects/obj_ncombat/Alarm_5.gml
```diff
@@ -330,17 +330,15 @@ if (defeat=0) and (reduce_power=true){
...
             power_reduction = 2;
-            power_fought = enemy_power;
         }
         new_power = enemy_power - power_reduction;
         new_power = max(new_power, 0);
 
         // Give some money for killing enemies?
-        var reward_table = [0, 5, 10, 20, 40, 80, 160, 320];
-        requisition_reward = reward_table[power_fought];
+        var _quad_factor = 8;
```
EttyKitty, line 340:
```suggestion
        var _quad_factor = 6;
```

Blogaugis (approved):

PR #653, opened 2025-03-31
feat: Full Livery options for companies

OH296:
### Purpose of changes
- companies can be assigned separate full livery settings that layer over the top of role livery settings
- marines can have their own personal livery 
    - livery order is as follows, personal > company > role > chapter

deathwatch marines will have deathwatch shoulder
deathwatch marines will have a custom colour right shoulder
next step is to to be able to tweak the priority order e.g favour role data or favour company data


objects/obj_garbage_collector/Step_0.gml
```diff
@@ -2,7 +2,7 @@ if (gc_timer > 0) {
     gc_timer -= 1;
     // show_debug_message($"obj_garbage_collector: gc_timer = {gc_timer}");
 } else {
-    gc_timer = 50; // Default is every frame, so de-facto 1;
+    gc_timer = 3; // Default is every frame, so de-facto 1;
```
EttyKitty, line 5: Have you tested this? The reason for why we created a manual GC timer, is for it to not run every frame, because it was corrupting arrays otherwise. I don't remember how small were values I tested, but at some point YYC started corrupting everything again.

EttyKitty: Yeah, sure. Could work.

PR #676, opened 2025-04-03
fix: non-crash error with tau fleets

VanWeapon:
### Purpose

- non-crash error showed up while i was doing save/load testing. 

### Testing done

- reloaded save where error was occurring and now it doesnt


EttyKitty:
I merged it already, so you're good. #673

PR #683, opened 2025-04-05
feat: Sternguard and Vanguard > Veteran; Loadout adjustments

EttyKitty:
### Purpose

- The idea behind Sternguard and Vanguard is cool and lore friendly, but in my opinion just doesn't work properly atm. 
- What you get, are weird, self assigning units, with no logic behind the assignment.
- I don't think that having them right now, in their current state, outweighs their clunkiness, and I can't bother with properly solving that, as I personally don't need it that much.
- I do think that having them merged, **for now**, is better than maintaining the status quo, and hoping that one day someone will repair them.
- The removal/addition is a couple lines of code, so nothing is really lost.

#### Vanguard and Sternguard

- Vanguard and Sternguard squads are converted into normal Veteran squads.
  - The loadout of these new squads is very similar to Sternguard loadout, consider these to be "Tacticals+".
  - If the player wants, they are free to change loadouts of their veteran squads and assign them to different columns.
- What the player loses factually:
  - A bunch of melee starting equipment that Vanguard spawned with.
  - Vanguard and Sternguard role names.
  - Nothing else I can think of.
- What the player gets:
  - No more headache with units jumping into random squads.
  - Ability to manually make their veterans into devastator like squads or assault like squads, or leave the base "tactical" like loadout.
#### Loadout changes
- Adjusted Salamander specific assault loadout to only replace the assault marine equipment. The whole thing is silly, and I see no reason to have it in its current state tbh.
- All other changes to weapon lists can be seen in the diff, I can't list them. Generally, I just tried to stop 1:1 adhering to the TT rules, as they are borderline retarded in some loadout choices.

#### If someone will want to properly repair Vanguards and Sternguards
- They are free to do so after these are disabled.
- First, columns of these two should be separate, to allow proper distinguish between the melee and ranged units.
- Second, there should be proper logic as to who gets attached to what squad. Not sure if separate roles are needed or not.
- Third, the player should have full control over the above logic.
- Otherwise, it will still be a good riddance.

### Testing done

- Launched the game as UM multiple times while testing and before creating the PR.

- None.

### DIFF
--- a/objects/obj_popup/Mouse_50.gml
+++ b/objects/obj_popup/Mouse_50.gml
@@ -321,7 +321,7 @@ if (point_in_rectangle(mouse_x, mouse_y, xx+1465, yy+499,xx+1576,yy+518)){// Pro
         variable_struct_set(role_squad_equivilances,obj_ini.role[100][9],"devastator_squad");
         variable_struct_set(role_squad_equivilances,obj_ini.role[100][10],"assault_squad");
         variable_struct_set(role_squad_equivilances,obj_ini.role[100][12],"scout_squad");
-        variable_struct_set(role_squad_equivilances,obj_ini.role[100][3],"sternguard_veteran_squad");
+        variable_struct_set(role_squad_equivilances,obj_ini.role[100][3],"veteran_squad");
         variable_struct_set(role_squad_equivilances,obj_ini.role[100][4],"terminator_squad");
 
         for(i=0;i<array_length(obj_controller.display_unit) && mahreens<500;i++){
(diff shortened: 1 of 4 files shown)


Blogaugis (approved):

KRdaMystic:
I think you could nuke the salamander's assault code as well its kinda just the regular loadout atp, it was better conceptually than it is now.

EttyKitty:
> I think you could nuke the salamander's assault code as well its kinda just the regular loadout atp, it was better conceptually than it is now.

It was always weird how only their assaults get a different loadout. I'll probably find some time, before merging this, to add a bunch more units into the code block, with more melta and flame stuff.
Going to be a hell to work with, considering how these data structures operate, though.

EttyKitty:
> I think you could nuke the salamander's assault code as well its kinda just the regular loadout atp, it was better conceptually than it is now.

Sorry, I take that back. I'm not making custom loadouts for Salamanders. I just don't want to deal with the copy-pasting of huge code chunks, and pinpointed access requires array giggling + struct.
No, not dealing with this. I've had enough for now.

PR #684, opened 2025-04-05
feat: Allow more column options for battleline squads

KRdaMystic:
Allow squads widely variable loadouts to be put in different formations. 

- Self-descriptive.

- Self-descriptive.

- Possibly making formation options a general location vs role specific eg (Row 1, 2, 3, & 4) vs (Assault, Tactical, Devastators, Scouts)

### Testing done

- None and I understand the risks. This is a very minor change that wont break anything

- None.

### DIFF
--- a/scripts/scr_initialize_custom/scr_initialize_custom.gml
+++ b/scripts/scr_initialize_custom/scr_initialize_custom.gml
@@ -1787,7 +1787,7 @@ function scr_initialize_custom() {
 			}],
 			["type_data", {
 				"display_data": $"{roles.devastator} {squad_name}",
-				"formation_options": ["devastator"],
+				"formation_options": ["devastator", "tactical", "assault", "scout"],
 			}]
 		],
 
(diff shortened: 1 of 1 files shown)


EttyKitty:
tbf, if you want, you can add even more options. they don't matter anyway.
it's only important to have the first one set right, as formations are assigned based on it, at the start of the game, iirc.

I hope to scrap this array property from all squads and just allow you to deploy any squad in any column, as I see no reason not to.
then just have a single formation specified, the one the squad will spawn with.

EttyKitty:
@KRdaMystic, is it okay if I push to this branch the full removal of the arbitrary formation pick restrictions from the squad structure?

EttyKitty:
but... nah. it's okay. not gonna. just gonna allow assaults to pick formations as well, because what if you want to.

EttyKitty:
@coderabbitai review

EttyKitty:
@coderabbitai, create an issue:
To remove the arbitrary column choice restriction from squads, and only have one column specified, as the default choice that the squad spawns with.

PR #685, opened 2025-04-05
feat: Fully working Omnissian Axe and repair specials

KRdaMystic:
### Purpose
- Make techmarines use the newly added Omnissian axe since its their iconic weapon

- Self-descriptive.

- Adding an acquisition process following a TechMarine's training that isn't purchase or forging.
### Testing done

- None and I understand the risks.

- None.


Blogaugis:
I think `Omnissian axe` is a bloat item and should not be in game.
`Power axe` should be enough. If we want even more debloating, we can combine power weapons into `power weapon`.

Blogaugis:
Otherwise, `Omnissian Axe` should provide repair capabilities and stat buffs to the user, if we want to do justice to the item.
Arguably, it should also be available in at least `master-crafted` equipment level.

KRdaMystic:
> I think `Omnissian axe` is a bloat item and should not be in game. `Power axe` should be enough. If we want even more debloating, we can combine power weapons into `power weapon`.

Frankly I disagree, This is similar to how we have the `Crozius Arcanum` for Chaplains in addition to the Force Staff, Axe and Sword for Librarians.


> Otherwise, `Omnissian Axe` should provide repair capabilities and stat buffs to the user, if we want to do justice to the item. Arguably, it should also be available in at least `master-crafted` equipment level.

I can agree with the possibility of bonuses to repair and possibly forge output but not in this PR.

CassieCaster (approved):
I strongly, massively disagree with Blogaugis. More over I massively support this course of action.

Blogaugis:
> Frankly I disagree, This is similar to how we have the `Crozius Arcanum` for Chaplains in addition to the Force Staff, Axe and Sword for Librarians.

You added the `Crozius Arcanum` yourself.
Not sure about the 2 other weapons for librarians (staff is the only one they had back in 0.65~).

Well, good luck adding the rest of `office weapons` for the remaining blokes in game.

Blogaugis (approved):
Since `Omnissian Axe` now actually serves a repair purpose, I now approve of this change.

EttyKitty:
@coderabbitai, create a TODO issue.
To rework the vehicle and, maybe, the marine recovery system. From being a predictable "if you have enough recovery points, you'll always recover all vehicles", to have a random chance of recovery, that increases with the techmarine skill. And items, instead of giving more recovery points, give bonuses to passing the random check.
Not sure yet how to implement this properly in relation to having multiple techmarines and other nuances, but will still better than what we have right now, as right now you can cheese this recovery mechanic.

EttyKitty:
@coderabbitai, create another issue:
To transfer all existing hardcoded values (that affect something in the gameplay), specific to weapons, armor and gear, from across the codebase to the new specials property of items.
Like I did with the `combi_tool` special, instead of it being hardcoded to check item names.

PR #711, opened 2025-04-09
fix: ai_b tau fleet scope

OH296:
fix of a fix whereby the game tried to access the characteristics of a tau fleet before ensuring the tau fleet existed


EttyKitty:
Wait, why does this matter? Expressions are evaluated from left to right. If flit == none it stops before checking the tau fleet, no?

EttyKitty:
so,
```
if (flit != "none"){
  if (flit.owner == eFACTION.Tau) {
```
should be the same as:
``` if (flit != "none" && flit.owner == eFACTION.Tau) {```
I understand why it wouldn't work if it was `||`.

EttyKitty:
I just know, that I've used this a lot of times throughout the code and never saw a problem.
Did you get a crash?

PR #720, opened 2025-04-11
feat: Weapon display data driven and Cthonian culture pack

OH296:
- make display data for weapons data driven and integrated into the complex system
- weapons can now have variants and even different holding positions for the same weapons
- weapons have access to all the visual options of standard complex components including culture options and exp and ranks based options allowing granular control of different visual types of weapons
- add new "shin" areas to marine sprites to allow morevariants
- create cthonian culture set sprites.
- start to implement bare head sprites (under the hood)

### DIFF
--- a/objects/obj_garbage_collector/Step_0.gml
+++ b/objects/obj_garbage_collector/Step_0.gml
@@ -3,7 +3,7 @@ if (gc_timer > 0) {
     // show_debug_message($"obj_garbage_collector: gc_timer = {gc_timer}");
 } else {
 
-    gc_timer = 50000; // Default is every frame, so de-facto 1;
+    gc_timer = 50; // Default is every frame, so de-facto 1;
 
     gc_collect();
 


scripts/scr_culture_visuals/scr_culture_visuals.gml
EttyKitty, line 1: All I said relates to this file.

scripts/scr_culture_visuals/scr_culture_visuals.gml
```diff
@@ -1781,3 +984,794 @@ function add_purity_seal_markers (){
...
+      position: "weapon",
+      weapon_data: {
+        display_type: "shield",
+        sprite: spr_weapon_storm2,
+        arm_type: 2,
+        ui_spec: true
+      },
+      body_types: [0, 1, 2]
+    }
+  ],
```
EttyKitty, line 1543: This is much better, but I still think that this is omega verbose.
`weapon_map` also looks obsolete now.

EttyKitty:
> The visual display data for the weapons has absolutely nothing to do with the weapons in game stats.

In my eyes, it's not really a question of whenever it has anything to do or not. I view the exact weapon as an object. All data related to the object is supposed to be attached to the object.
If you do that in multiple places, yes, you may make it easier to read through and reduce the file size, but is a mess imo.

Having data about one object split around multiple files, imo, is bad. I would much rather have ALL possible data in the code that is related to each weapon attached to the weapon, than anything.
I come from modding [this](https://www.ufopaedia.org/index.php/Ruleset_Reference_Nightly_(OpenXcom)#Items), and while imperfect, I never had problems with items having *too many* property fields. If anything, I've had difficulties with people splitting item data across multiple files, and me having to search for it all around the project. And I've had A LOT of experience with this question.

If the concern, again, is object property bloat and file size bloat, reducing the amount of copy-pasted visual data may make more sense.
A lot of data I see in your example is copy-pasted stuff, that takes a lot of place and often the same for many weapons.
Just put repeated things into a template (`big_shield`) and apply it to weapons, like so:
```js
  "Storm Shield": {
      draw_type: "big_shield",
	  sprite_variants: [
	      {
	      cultures: ["knightly"],
	      sprite: "spr_weapon_storm"
	      },
	      {
	      sprite: "spr_weapon_storm_tall"
	      },
	      {
	      sprite: "spr_weapon_storm_tall_2"
	      },
	      {
	      sprite: "spr_weapon_storm2"
	      }
	  ]
    }
    
    ////////////////
    
      "big_shield": {
            position: "weapon",
		    weapon_data: {
		      display_type: "shield",
		      sprite: spr_weapon_storm_tall_2,
		      arm_type: 2,
		      ui_spec: true
		    },
		    body_types: [0, 1, 2]
    }
```

Blogaugis:
So, wait, if I were to implement tech tree expansion of tribal-iron-steel-future technology related weapons, I'd now need to also create several sub-variants of these weapons?

EttyKitty:
> So, wait, if I were to implement tech tree expansion of tribal-iron-steel-future technology related weapons, I'd now need to also create several sub-variants of these weapons?

Variants are not required.

scripts/scr_culture_visuals/scr_culture_visuals.gml
EttyKitty: > It is verbose yes but it's also really accurate, i specifically know everything is formatted exactly how i want and i don't have to come back because of minor, format changes.

I don't understand what "accurate" means in this case.
You're copy-pasting the exact same fields with the exact same values. That's not accurate, that's just a violation of DRY, line bloat and a maintenance hell, for no good reason imo.

> not really happy with this as i specifically want to have the option of having varient grip and hand positions on a single weapon as you say number of data fields is not an issue

And you're free to do so, by just adding the needed fields with custom values to each variant that needs it (just rename the field to better reflect what it does), while having default values inherited from the base `draw_type`, and overwritten case by case.
Or by creating a separate `draw_type`, if there are more than 1 case of items that use the same override.
That's one of the reasons why we have constructors - create a template object, with default values, and then edit it to shape the form you require. Can also argue that we can delete them, and manually create each struct to have full control, and for it to be 100% "accurate", as you say.

EttyKitty:
> I've never really understood what this means, even on one small screen keeping multiple screens viewable on any competent text editor is pretty easy. frequently have in excess of 15 files open at any one time generally having two or more up side by side or having a single file in split view, to me it will always make far more sense to keep data in manageable packets for quick reference.

I can't say anything to this, as I never work with two files open, unless I actually have to compare something, or copy, or there is just no choice. I would never prefer this, if I have an option.

Even having my whole screen dedicated to work with just one file sometimes feels not enough, with all the additional stuff on the scree, aside from the file.
I have two 1920x1080 monitors tho, but I never open two files at the same time to work with, if I can help it. The other screen is taken for something else.
The usage of split view, multiple monitors, big monitors, etc., should never be expected of people who work with the code, as some kind of required default rule, without which the code is unreadable.
Up until now, I've used a 1280x1024 monitor, and I couldn't work even with one file properly.

And for quick reference, it also makes more sense to have all data related to the object in one place, as that's just one exact pocket - a pocket with all object data.
If I search throughout the project for "Storm Shield", I want to find exactly one place where all of its data is stored, and however many where it's used.
If a modder comes to mod the item, he'll want one json file with all related item data. Not 7 billion separate ones, for combat data, shop data, visual data, ass data, rarity data, reward data, artifact data, material data, research data.
If he wants to add a new item, he'll want to copy-paste an existing item in one place, and edit it to become a new one. Not do that 7 billion times in all other files.

EttyKitty:
But creating various templates, presets and other shit, to reduce the number of fields on each object - is totally fine. Like the example with `draw_type`, which allows to reduce the number of fields by x5 times or so.

EttyKitty:
> Which i don't like and find even more cumbersome and annoying i will go with it if it's more to your preference but imo it just means i end up stacking a shit load of ablative and pointless code into the evaluator

Just from the top of my head.
```js
var weapons = {
    "Assault Cannon": {
        "abbreviation": "AssCann",
		"attack": {
			"standard": 56,
			"master_crafted": 56,
			"artifact": 56
		},
		"description": "A heavy rotary autocannon with a devastating fire rate that can be counted in the hundreds per minute. It is incredibly effective against infantry and light armored targets.",
		"melee_hands": 2.1,
		"ranged_hands": 3,
		"ammo": 20,
		"reload": 3,
		"range": 16,
		"spli": 10,
		"arp": 12,
		"tags": ["heavy_ranged", "dreadnought"],

        // with a default and variatnts
        "display_datav1": {
            "draw_type": "terminator_cannon",
            "sprite": spr_weapon_assca,
            "variants" : [
                {
                    sprite: spr_weapon_assca2,
                    "draw_type": "heavy_cannon",
                    conditions: [
                        {
                            "unit_type": "Power Armour"
                        }
                    ]
                }
            ],
        },

        // or without a default:
        "display_datav2": [
            {
                "draw_type": "terminator_cannon",
                "sprite": spr_weapon_assca,
                "conditions": [
                    {
                        "unit_type": "Terminator"
                    }
                ]
            },
            {
                "draw_type": "heavy_cannon",
                "sprite": spr_weapon_assca2,
                "conditions": [
                    {
                        "unit_type": "Power Armour"
                    }
                ]
            },
        ],

        // or shorter, but with will require condition sets prebaked:
        "display_datav3": [
            {
                "draw_type": "terminator_cannon",
                "sprite": spr_weapon_assca,
                "conditions_set": "terminator_unit",
            },
            {
                "draw_type": "heavy_cannon",
                "sprite": spr_weapon_assca2,
                "conditions_set": "normal_unit",
            },
            {
                "draw_type": "heavy_cannon",
                "sprite": spr_weapon_assca3,
                "conditions_set": "kightly",
            }
        ],

        // or even better for a quick dictionary lookup, if you know what to search for:
        "display_datav4": {
            terminator_unit: {
                "draw_type": "terminator_cannon",
                "sprite": spr_weapon_assca,
            },
            normal_unit: {
                "draw_type": "heavy_cannon",
                "sprite": spr_weapon_assca2,
            },
            kightly: {
                "draw_type": "heavy_cannon",
                "sprite": spr_weapon_assca3,
            }
        },
    }
};

// A quick example
var draw_types = {
    "terminator_cannon": {
        arm_type: 1,
        hand_type: 0,  
        body_types: [2]  
    },
    "heavy_cannon": {
        arm_type: 2,
        hand_type: 1,  
        body_types: [1]  
    }
};

// A quick example
var condition_sets = {
    terminator_unit: {
        "unit_type": "Terminator"
    },
    normal_unit: {
        "unit_type": "Power Armour"
    },
    kightly: {
        "culture": "Knightly"
    }
}
```

EttyKitty:
> in particular what is the purpose of this?
> 
> > ```
> > kightly: {
> >    "culture": "Knightly"
> > ```
> like this is just a layered way of doing the same thing

That's just an example. You can add additional conditions to it, i.e. "unit_type: "space_marine", or rename and separate into multiple knightly variants, i.e. for terminator and normal units.
You can name it "OMEGA KNIGHT" but leave the culture condition the same. Then just use "OMEGA KNIGHT" in the condition_set field.
Or leave as is, because the main purpose is just to just have a quick lookup and a standardized set of `condition_sets`, even if names match. 

If you remove it and argue that you can just match the culture from the getgo, directly by the `condition_set` name, are you're going to try and lookup every `condition_set` value in the culture list? That's just overhead for no reason. 
Adding a separate field in addition to the `condition_set`, i.e. `culture` is just doing the same job the condition set does, just because you don't want condition sets with the same name as the condition?
Take the `condition_set` from the item, immediately, without any loops or checks, do a function call to check if all conditions match, i.e. `conditions_check(condition_set)`, get true or false - done. Bonus points if squeeze the priority logic somewhere there.

I'm not going to pretend that my examples are perfect and finished, but the logic I'm trying to convey is there.

> I can see no way that this isn't about a million ways more unnecessary

I don't know what to say at this point.
If retyping the same exact values with no change 10 times for 100 items, and/or having 7 billion nested structs and arrays everywhere is better than, for example, the 4th variant where you basically just use prebaked templates that just do the same thing with minimum retyping and nesting - I can't really say anything to prove otherwise.

> did you see this?

Yeah, I did. I don't really care for moving it rn or as soon as possible.
I just don't want to leave it as it is forever, because I, personally, see it as a bad structure and may just start avoiding the whole thing in the future.
Likewise, I have no idea if other people agree with this view, or if only I see it this way, so I don't want to sway you too much, to be honest.

EttyKitty:
I also want to just say this, as this is my view:

Every data structure that is possible to extract to JSON and let people modify/see, should be extracted to JSON.
Every JSON that people can modify/see, should be as simple as possible, so that a 3 years old can modify it and understand how it works without a 300-page-long manual. Or it's hard to look at, thank you, bye.
Every attempt at modification, should be as simple, quick, and not tedious as possible. Minimum nesting, minimum retyping, maximum reusable modules/templates. Or it's a bother, thank you, bye.
Every maintenance of an old modifications/code should be the same as the two above. Or the code starts to break, and you have to travel 8k lines to find what was not updated, across copy-pasted values and fields, that don't need to be there but are - not worth the effort, let it die, bye.

Otherwise, you lose a lot of people who could contribute, but don't because either they can't understand what is going on, or because they don't want to as it's just tiresome.
Or a lot of contributors, who avoid areas of code, because it's just a bother to look at/interact with.

EttyKitty:
> Ok i'm not really sure where to go here, i know there are optimisations to make, but i'm not really sure how to implement them in a desired way as opposed to my way. so i'm just gonna mothball this pr for a while till i can think of something

No need tbh. If it works - it works. Just merge it as it is. Stuff can be improved/refactored later.
Imo it's more important to get the feature in, to avoid future branch conflicts and discouragement.

EttyKitty:
Also, is it intended for the head stuff to be in this PR?

EttyKitty (changes requested):
I'm not going to stall this for longer, as stuff can be improved later.

The feature works and that's good enough.

Just pick a good name for the PR, as it does a lot of stuff aside from the weapon visual data.

EttyKitty (approved):

PR #758, opened 2025-04-22
build: Disable deprecated options

EttyKitty:
### Purpose

- These probably produce unintended behavior and should be deprecated at some point.

- Disable all of them.
image

- Nothing.

### Testing done

- Launched the game - nothing seems broken. But there is a very high chance that something will break.

- None.


MCPO-Spartan-117:
i really doubt CoW even works.

EttyKitty:
> ngl i think disabling copy on write this far through is a bad idea. It will almost certainly break a load of things that used an array for processing and we don't really need it at this point

No idea. I created this PR just for the experiment. I also think that something may break, but I didn't see anything break explicitly at first glance.

PR #763, opened 2025-04-23
chore: Star name draw surface removal

OH296:
### Purpose
 - draw star surfaces by storing sprites instead of surfaces, add a cleanup event to obj controller to ensure
   - name generator gets rebuilt stopping star names from running out
   - sprites and star draw data gets cleared from memory properly
- use a ds map for storing star draw data as its a critical component and we now know thery are much quicker to search

- Self-descriptive.

- Nothing.

### Testing done

- None.

### DIFF
--- a/objects/obj_controller/CleanUp_0.gml
+++ b/objects/obj_controller/CleanUp_0.gml
@@ -8,3 +8,5 @@ for (var i=0;i<array_length(_star_arrays);i++){
 		sprite_delete(_star_arrays[i]);
 	}
 }
+
+ds_map_destroy(obj_controller.star_sprites);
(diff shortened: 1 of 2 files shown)


objects/obj_controller/Create_0.gml
```diff
@@ -98,6 +98,9 @@ obj_cuicons.alarm[1]=1; // Clean up custom icons
 map_scale = 1;
 scale_mod = 1;
 
+star_sprites = ds_map_create();
```
EttyKitty, line 101: Two questions: is it okay to never destroy it? And is it okay with save/loading?

EttyKitty: > Destroys with obj_Controller cleanup automatically to avoid issuesand yes fine with save loading or. should be i'll run a couple of tests

I think ds_ and other stuff is not destroyed with an object. You destroy it explicitly, with a function, or it stays.

EttyKitty (approved):

PR #765, opened 2025-04-25
build: Surface changes; Texture Group changes; Asset regrouping

EttyKitty:
### Purpose

- Fix memory leaks and leftover surfaces stuck in the memory.
- Have textures in proper groups and not just one (still not ideal).
- Make asset grouping a bit better.

- Rename `clear_and_free_surface` with `surface_clear_and_free`, to be a bit more standardized.
- Replace the remaining `surface_free` calls with `surface_clear_and_free`.
- Create a couple of texture groups, to separate texture pages a bit. The "Unused" group is also dynamic.
- Move assets/folders around in the GM asset view, to hopefully group them a bit more intuitive.

- Texture groups are still not ideal and there is still a bunch of texture swapping, and currently unneeded sprites on one page.
- Unused texture group probably has some sprites that are still used, making it load into the game even if it's dynamic.
- There are probably more unused sprites, that don't need to be in the main groups and just eat RAM.

### Testing done

- Launched the game a lot of times to test each change.

- None.


scripts/scr_ComplexSet/scr_ComplexSet.gml
```diff
@@ -794,7 +790,7 @@ function ComplexSet(_unit) constructor{
             _complex_helm = _comp_helms.veteran;
         }
         if (is_struct(_complex_helm) 
-            && struct_exists(self, "head") && draw_helms){
+            && struct_exists(self, "head") && draw_helms && !sprite_exists(head)){
```
EttyKitty, line 793: Wait, I only now thought, this will make it so that helm sprites don't get updated.

EttyKitty: I guess I should just delete the old sprite instead, if exists.

EttyKitty: Done.

PR #817, opened 2025-05-13
refactor: obj_controller and ui_manage temp array stuff

EttyKitty:
### Purpose

- Maybe fix a bug.
- Fix the array initialization to be sane and let the IDE know about it.
- Make the code a tiny bit more readable.

- Initialize temp array with `array_create` and (new) `array_set_range`, and fix a weird use of `i` in a `t` based loop.
- Assign `temp[120]` (selected unit) to a local var where it's used.
- Replace a weird `cn` var with just using the full object name.
- Add `obj_controller` accessor where temp vars are used to explicitly tell where they are coming from, as otherwise IDE doesn't like them.

- Nothing.

### Testing done

- I HAVEN'T TESTED THIS, BECAUSE I CAN'T ATM!

- Bug report: https://discord.com/channels/714022226810372107/1371540899649290492

### DIFF
--- a/objects/obj_controller/Create_0.gml
+++ b/objects/obj_controller/Create_0.gml
@@ -750,14 +750,11 @@ trade_chip=0;
 trade_info=0;
 zui=0;
 // Variables for management
-for(var t=0; i<200; t++){
-    temp[t]="";
-}
-temp[90]=0;
-temp[9000]="";
-for(var t=4700; t<=4719; t++){
-    temp[t]=0;temp[t+100]=0;
-}
+var array_size = 9001;
+temp = array_create(array_size, 0);
+array_set_range(temp, 0, 199, "");
+temp[90] = 0;
+temp[9000] = "";
 // ** Resets all audiences **
 audiences=0;
 audien[0]=0;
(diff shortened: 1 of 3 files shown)


EttyKitty:
> Long term we should shift to a struct so we can have `temp_struct.tooltip_text = ""` instead of doing `tooltip_text = obj_controller.temp[103];` i guess? Unless there is a good reason for these arrays?

There is no good reason for these arrays. I hate them and hope they die.
You can avoid the use of struct, and just have a separate temp object with all of these things as its instance vars. Not super important here, but less nesting. So it's just `obj_temp.tooltip_text`, instead of `obj_controller.temp_struct.tooltip_text`.

EttyKitty:
> > > `obj_temp.tooltip_text`, instead of `obj_controller.temp_struct.tooltip_text`.
> 
> does occur to me though the whole thing runs within the scope of obj_controller anyway so why does any of this specify obj_controller in the first place?

Because my IDE (vscode) can't catch the reference otherwise, for some reason, and just lights up the whole file red. I assume I can use some kind of tag to fix that, but I saw that the accessor was already used in some places here (through the old cn var), and decided not to bother further and just carpet bomb it.

I also for some reason assumed that it *can* run from some other scope, because of the original cn var.

Totally fine to nuke all of these in one go, tho, if they don't matter.

PR #862, opened 2025-06-05
feat: New Opening Game text with servo skull

OH296:
### Purpose

- Self-descriptive.

- Self-descriptive.

- Nothing.

### Testing done

- None.


EttyKitty:
An alternative route - you re-do half of the asset, so that it's a perfect base for the text. Not sure how far you want to go with this, though, and not sure if it'll be worth it in the end.

PR #890, opened 2025-06-23
feat: New loot table concept

OH296:
## Purpose and Description

- Self-descriptive.

## Testing done

-


sa201213:
Maybe different Xenos species would have different drop ratios? E.G. it would be easier to loot Tau gear (just advanced tech) than to loot Eldar gear (advanced _and_ works on psionics, so practically unusable for non-psykers).

In-canon, Ork gear is complete wildcard when it comes to looting it: sometimes it works normally, and sometimes it outright refuses to work when wielded by any non-Orks due to requiring WAAAAGH energy.

PR #918, opened 2025-07-13
fix: Scout right pauldron color fix

Werwolf-12:


ChapterMaster.yyp
EttyKitty, line 1: Are changes to this file intended?

PR #939, opened 2025-07-21
fix: Obliterated Fleet Comp Resend

CptMacTavish2224:
## Changes and Fixes
Obliterated with Kings of Space and/or Boarders will now receive their bonus vessels on start

## Testing done
Fully tested


EttyKitty:
Merged in #940.

PR #951, opened 2025-07-29
fix: crash from non-equip-able artefacts

OH296:
## Purpose and Description

- Self-descriptive.

## Testing done

-


Blogaugis (commented):
Might want to support changing department names at some point.

PR #952, opened 2025-07-29
fix: Combat Log

Blogaugis:
## Purpose and Description
Fixes misleading statements about no damage - damage is dealt, and marine/hireling/vehicle(?) will lose hit points. It now should state if marine/hireling/vehicle(?) is downed (no longer an active combatant) or not.

## Testing done
- None, as it is a small and contained flavor text tweak.

https://discord.com/channels/714022226810372107/1398987282463068270 - bug report.


Blogaugis:
@EttyKitty As far as I'm aware, it is ready to be merged.

PR #1004, opened 2025-09-06
fix: highlight and shadow and multi-part weapons

OH296:
## Purpose and Description

- Self-descriptive.

## Testing done

-


ChapterMaster.yyp
```diff
@@ -576,7 +576,7 @@
   "isEcma":false,
   "LibraryEmitters":[],
   "MetaData":{
-    "IDEVersion":"2024.1400.0.815",
+    "IDEVersion":"2024.1400.0.884",
```
Blogaugis, line 579: Which IDE version are we supposed to use?

We may not be able to reproduce bugs if our IDEs are different.

PR #1028, opened 2025-09-28
fix: refactor of scr_en_weapon.gml to switch cases

wfelipeagudelo:
## Purpose and Description

- Changes and improves the handling of conditions to be switch case instead of dozens of if conditions

## Testing done

- 20 turns
- multiple ground combat batlles (vs orks, genestealers)
- faster load for all combat actions

- None


EttyKitty:
> If its only stats you changed it should be simple, mainly its pure switch statements now, other than that not much changed in file.
> 
> I could help with the merge in ur branch worst case

yeah, I think that there shouldn't be any bad conflicts. the branch is here #439, if of any importance.

in any case, I can't tell when it'll be finished, so I see no reason to stall this PR, even if there will be conflicts.

PR #1052, opened 2025-11-05
fix: hardcoded values counting vehicles (and marines) on managmenet pane.

swang300:
…management screen, hopefully

## Purpose and Description

- Vehicles were not being counted correctly on the management screen.  fixed
- also, fixed some hardcoded values to check array length.

## Testing done

-


OH296 (changes requested):
one minor change otherwise all seems good to me, as a note we generally are trying to avoid the use of `then` and are swapping it out with the more conventional `if (condition){<do thing>}` format but that is a a lesser concern that i'm not going to enforce minor changes

OH296:
@coderabbitai review

OH296 (approved):

PR #1056, opened 2025-11-11
fix: plasma pistol ammo bug

Werwolf-12:
## Purpose and Description

- Self-descriptive.

## Testing done

-


OH296:
when you start a new piece of work you need to create a new branch off of main as we `quash` our commites when merging to main. 
So waht you need to do is create a new branch off main and then cherry pick `palsma pistol FIX` and `Ryza Plasma gun fix`.

I can handle this for you if you like and can explain this more if needed but i thought it would be better for yo to read this and at least try get your head round it rather than steamrolling around you

PR #1236, opened 2026-06-15
refactor(portability): normalize path separators to forward slashes for Linux compatibility

RedMarcher:
#### What does this PR do?
* **Path Separator Refactoring**: Converts hardcoded Windows-style double-backslashes (`\\`) in GML path strings to cross-platform forward slashes (`/`) across 16 files in the `scripts/` and `objects/` directories.
* **Universal Compatibility**: Standardizes pathing so files resolve natively on Linux and macOS environments (which strictly use `/`), while maintaining full support on Windows (which natively accepts `/` at the system level).
* **OS-Aware Path Presentation**: Updates the debug/error log saver logic in ErrorHandler.gml to dynamically detect the runner OS (`os_type`), using backslashes for presentation on Windows and forward slashes on Linux.
* **Avoids String Escaping Warnings**: Eliminates GML warning/parsing issues associated with backslash escaping (`\\`) inside standard string literals.
#### Why are these changes needed?
Hardcoded backslashes are a major blocker for native Linux builds, as Unix-like file systems do not recognize backslashes as directory separators. Using the forward slash is the industry-standard way to write platform-agnostic file paths in GameMaker (and general C++ file operations).
* **Reference Documentation**:
  * [GameMaker Manual - The File System](https://manual.gamemaker.io/monthly/en/Additional_Information/The_File_System.htm) (Relative paths using `/` are automatically resolved sandboxed across all targets).
  * [Microsoft Win32 File Paths Documentation](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats) ("Windows also recognizes the alternative directory separator character, which is the forward slash (`/`)").
#### How was this tested?
**Verification of Strings**:
   * Analyzed the exact differences using `git diff` to confirm only filesystem paths were modified.
   * Verified that escape sequences (such as `\n` and `\t`) and literal backslashes in actual dialogue text (such as `\\\\Civil War\\\\` in scr_dialogue.gml) were safely ignored by the refactoring tool.

---
## Summary by cubic
Standardized all file path strings to use forward slashes to enable native Linux/macOS builds and remove GML backslash-escape issues. Error log messages are now OS-aware while keeping full Windows support.

- **Refactors**
  - Replaced `\\` with `/` in path strings across `scripts/` and `objects/`.
  - Updated path macros and directory creation in `scripts/__init/__init.gml` to use `/`; switched all JSON, INI, and image loads accordingly.
  - Error handler now presents `Logs\` on Windows and `Logs/` on Unix, while using `/` internally.
  - Adjusted image loaders, including `scr_image_cache`, to build paths with `/` and removed backslash conversions.
  - Avoids string-escape warnings in GML and unblocks Linux builds without changing asset locations.
  - No gameplay or data changes; Windows continues to work with `/`.

<sup>Written for commit df8c500409d02a437919916ea2fa8609a8e88551. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Adeptus-Dominus/ChapterMaster/pull/1236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/objects/obj_event_log/Create_0.gml
+++ b/objects/obj_event_log/Create_0.gml
@@ -23,8 +23,8 @@ repeat (101) {
     e += 1;
     topics[e] = "";
 }
-if (file_exists("main\\help.ini")) {
-    ini_open("main\\help.ini");
+if (file_exists("main/help.ini")) {
+    ini_open("main/help.ini");
     var ch;
     ch = 0;
     repeat (100) {
--- a/objects/obj_event_log/Draw_0.gml
+++ b/objects/obj_event_log/Draw_0.gml
@@ -118,7 +118,7 @@ if (help == 1) {
                 draw_set_alpha(1);
                 if (mouse_button_clicked()) {
                     topic = topics[t];
-                    ini_open("main\\help.ini");
+                    ini_open("main/help.ini");
                     info = ini_read_string(string(t), "info", "");
                     strategy = ini_read_string(string(t), "strategy", "");
                     main_info = ini_read_string(string(t), "main_info", "");
--- a/objects/obj_ini/Create_0.gml
+++ b/objects/obj_ini/Create_0.gml
@@ -271,7 +271,7 @@ deserialize = function(save_data) {
     var all_names = struct_get_names(save_data);
 
     if (!array_contains(all_names, "chapter_squad_arrangement")) {
-        obj_ini.chapter_squad_arrangement = json_to_gamemaker(working_directory + $"main\\squads\\company_squad_builds.json", json_parse);
+        obj_ini.chapter_squad_arrangement = json_to_gamemaker(working_directory + $"main/squads/company_squad_builds.json", json_parse);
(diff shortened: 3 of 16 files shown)


OH296:
@EttyKitty  i think the gobo formatter may have added in the majority of the "//" if not the case all good but if is the case ay chance it can be removed so this does not get reverted down the line?

EttyKitty:
> @EttyKitty i think the gobo formatter may have added in the majority of the "//" if not the case all good but if is the case ay chance it can be removed so this does not get reverted down the line?

Nah. Gobo doesn't touch string contents.
image

PR #1423, opened 2026-08-03
fix: Decrement loop index after removing operative in cancel loop

Karlious:
## What's wrong

In `scripts/scr_company_struct/scr_company_struct.gml`, the squad-assignment
"Cancel Assignment" button handler removes the squad's entry (or entries)
from the target planet's `p_operatives` array:

```gml
var planet = _cur_squad.assignment.ident;
var operation;
for (var i = 0; i < array_length(cancel_system.p_operatives[planet]); i++) {
    operation = cancel_system.p_operatives[planet][i];
    if (operation.type == "squad" && operation.reference == _cur_squad.uid) {
        array_delete(cancel_system.p_operatives[planet], i, 1);
    }
}
```

This is a forward `for` loop that calls `array_delete` on the array it's
iterating without decrementing `i` (or breaking) afterwards. When two
matching entries for the same squad sit at adjacent indices, deleting index
`i` shifts the second match down into slot `i`, and the loop's `i++` then
steps straight past it — the second entry is never removed.

## Why this is clearly unintended

- `PlanetData.add_operatives` (`scr_PlanetData.gml:122-125`) pushes a new
  entry onto `p_operatives[planet]` unconditionally every time a squad is
  put on assignment, with no check for an existing entry for that squad.
  Combined with the fact that not every code path that clears
  `assignment` reliably removes the matching `p_operatives` entry first,
  duplicate or stale entries for the same squad are a realistic outcome of
  normal cancel/reassign cycles, not a hypothetical. This loop is the point
  in the codebase specifically responsible for cleaning such entries out on
  cancel, and it silently fails to do so completely whenever two matches
  happen to be adjacent.
- Independent of how duplicates arise, "delete matching elements from an
  array while iterating it forward" is a well-known correctness trap in
  any language with array-index semantics like this — GML's `array_delete`
  shifts every later element down by one, so failing to compensate the loop
  index is a defect in the loop itself, not something that depends on how
  the array got into a particular state. The loop's own structure (no
  `break`, intended to catch every match) shows it was written to handle
  more than one matching entry, which makes the missing index compensation
  a bug rather than an accepted limitation.
- `git log -S "cancel_system.p_operatives"` shows a prior commit
  (`1a2f73db7`) touched this exact loop to fix an unrelated scope bug but
  left this pattern untouched, with no comment suggesting the behavior is
  deliberate.

## Before / after behavior

**Before:** if a planet's `p_operatives` array ever has two adjacent
entries referencing the same squad, cancelling that squad's assignment
removes only the first; the second remains, so the squad stays
double-counted by anything that iterates operatives per planet (e.g.
garrison totals).

**After:** decrementing `i` after a successful `array_delete` re-checks the
shifted-down entry on the next iteration, so all matching entries —
adjacent or not — are removed. This makes the removal loop correct
regardless of how or whether duplicate entries arise elsewhere.

## How to observe in game

This is a delete-during-iteration defect in the removal loop itself, so it
only produces a visible skip when the `p_operatives` array already contains
two adjacent entries for the same squad at the moment "Cancel Assignment"
is pressed — a state that can arise from repeated assignment cycles, since
`add_operatives` never deduplicates on push. The fix makes the loop correct
independent of how such a state occurs: inspect
`cancel_system.p_operatives[planet]` before and after pressing "Cancel
Assignment" on a squad with two adjacent matching entries — before the fix,
one survives; after the fix, both are removed.

---
## Summary by cubic
Canceling a squad assignment now deletes the squad’s operative entry and exits the loop immediately. This avoids index-shift issues and extra iterations, keeping per-planet counts correct.

- **Bug Fixes**
  - Break out of the removal loop after `array_delete` in `scripts/scr_company_struct/scr_company_struct.gml` (a squad `uid` appears at most once).

<sup>Written for commit 14e0122debb8e7617bdd38305aade38990967955. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Adeptus-Dominus/ChapterMaster/pull/1423?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/scr_company_struct/scr_company_struct.gml
+++ b/scripts/scr_company_struct/scr_company_struct.gml
@@ -345,6 +345,7 @@ function CompanyStruct(comp) constructor {
                         operation = cancel_system.p_operatives[planet][i];
                         if (operation.type == "squad" && operation.reference == _cur_squad.uid) {
                             array_delete(cancel_system.p_operatives[planet], i, 1);
+                            break;
                         }
                     }
                 }


OH296:
easier to just break the loop as there can't be more than one squad with the same UUID

Karlious:
Done — switched to `break` as suggested. One observation for whenever it's useful: `PlanetData.add_operatives` pushes unconditionally, so cancelling and re-issuing the same garrison duty can leave more than one entry for a squad over time. With `break` a cancel removes one entry per click, which still converges — just mentioning it in case a dedupe on the add side is ever worth a look. Also stripped the stray drafts md from this branch.

PR #1424, opened 2026-08-04
fix: migrate marine specials from the legacy spe array

The-Real-Nyx:
Loading a save made before the `obj_ini.spe` parallel array refactor crashed as soon as you opened a Librarian's panel in the company management view. `specials` used to be a method reading `obj_ini.spe[company][marine_number]`, and `jsonify_marine_struct()` skips methods so old saves carry no `specials` key at all, while `powers_known` was a plain field and round-tripped fine. `psy_discipline()` then reads `_specials_array[0]` on an empty array throwing an error. Fresh saves post refactor do not create the issue and legacy saves will be non-breaking with the refactor after first load and stored non-breaking after first fresh save.

## Changes
- Add `migrate_legacy_marine_specials()` to `obj_ini` `deserialize`, restoring `specials` from the legacy `spe` array when present. Skips marines that already carry their own string, and is bounded by TTRPG so the padding rows past company 10 never reach `fetch_unit()` and thus can't create cascading errors if any legacy save put data in them.
- Guard `psy_discipline()` against an empty specials, throw an error when a unit has powers but no discipline. This pairing cannot legitimately occur, so the guard acts as catch in case the migration shim creates weird broken marines in some circumstance or else a marine somehow ends up with specials but no discipline. Marines with no discipline but no specials will just get gracefully ejected instead as the most likely case is an unguarded call to `psy_discipline()` created them and they present no issue regardless.
## Testing
- Loaded old and new saves side by side. The crash only reproduces on legacy saves without the migration shim, and the migration causes no observable issues.

---
## Summary by cubic
Fixes a crash when opening the Librarian panel on legacy saves by restoring marine specials from the old `spe` array. Also prevents errors when a unit has no specials.

- **Bug Fixes**
  - Migrate legacy `spe` data during deserialize to repopulate each marine’s specials; skip marines that already have a value and bound iteration to `TTRPG`.
  - Guard psy_discipline to handle empty specials and log when powers exist without a discipline, preventing out-of-bounds errors on old saves.

<sup>Written for commit 09fe0c417fb24322c9b079b74081b3f643349db2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Adeptus-Dominus/ChapterMaster/pull/1424?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/objects/obj_ini/Create_0.gml
+++ b/objects/obj_ini/Create_0.gml
@@ -319,6 +319,48 @@ deserialize = function(save_data) {
         }
     }
 
+    /// @desc Restores marine `specials` from obj_ini.spe, the parallel array it used to live in.
+    ///    Saves predating that move carry no `specials` key at all, as it was a method and jsonify_marine_struct() skips methods.
+    ///    `powers_known` was already a plain field and did round-trip, so without this those psykers load knowing powers with no discipline behind them.
+    /// @param {Array<Array<String>>} _legacy_specials The saved spe array, indexed [company][marine]
+    /// @returns {undefined}
+    function migrate_legacy_marine_specials(_legacy_specials) {
+        if (!is_array(_legacy_specials)) {
+            return;
+        }
+        var _migrated = 0;
+        var _company_count = min(array_length(_legacy_specials), array_length(TTRPG));
+        for (var _coy = 0; _coy < _company_count; _coy++) {
+            var _company_row = _legacy_specials[_coy];
+            if (!is_array(_company_row)) {
+                continue; // the legacy array is ragged; stray non-array rows carry nothing
+            }
+            var _row_length = array_length(_company_row);
+            for (var _mar = 0; _mar < _row_length; _mar++) {
+                var _legacy_string = _company_row[_mar];
+                if (!is_string(_legacy_string) || _legacy_string == "") {
+                    continue;
+                }
+                var _unit = fetch_unit([_coy, _mar]);
+                if (!is_struct(_unit)) {
+                    continue;
+                }
+                if (is_string(_unit.specials) && _unit.specials != "") {
+                    continue;
+                }
+                _unit.specials = _legacy_string;
+                _migrated++;
+            }
+        }
+        if (_migrated > 0) {
+            LOGGER.info($"Migrated specials for {_migrated} marines from the legacy obj_ini.spe array");
+        }
+    }
+
+    if (struct_exists(save_data, "spe")) {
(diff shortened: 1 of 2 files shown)


OH296:
ok so when i merge the full deletion of all parallel arrays what ii will do is augment this function to iterate through all the destroyed arrays and fill all of the new variables in the struct, this way this will allow for a full migration

OH296:
The changes are so braod and varied i'm afriaid i just don't think it's going to be possible to acomodate reverse compatibility however i've been having a bit of a play around and  thinking of implenting a migration sequence that will work a bit like db migration so each save logs the last commit it was connected too and runs all relevant migration scripts since that point if any relevent migrations are needed.

I will outline this when i have more of a plan but it should avoid these unfortunate occurrences happening

The-Real-Nyx:
> The changes are so braod and varied i'm afriaid i just don't think it's going to be possible to acomodate reverse compatibility however i've been having a bit of a play around and thinking of implenting a migration sequence that will work a bit like db migration so each save logs the last commit it was connected too and runs all relevant migration scripts since that point if any relevent migrations are needed.
> 
> I will outline this when i have more of a plan but it should avoid these unfortunate occurrences happening

Don't use commits for it. I mean I can't stop you but it's a bad way to do it. You want a monotonic int that tracks schema version so migrations are just an ordered list of pure functions that run in order on the deserialized inert data, which means the migrations prolly want to happen before obj_ini runs too.

But also I ain't the one making so do what your heart desires or whatever.

OH296:
> Don't use commits for it. I mean I can't stop you but it's a bad way to do it. You want a monotonic int that tracks schema version so migrations are just an ordered list of pure functions that run in order on the deserialized inert data, which means the migrations prolly want to happen before obj_ini runs too.

Ah ok appreciate the advice, other than db work i've never built a migration system like this before so any thoughts are great. I think given how our serialiser works the best place to slip in migrations would be after items are read but before ay custom mutations/parse methods take place but that would probably mean distributing migration methods across the code to each position where information is parsed into each object pro rata e.g obj_ini, obj_controller, obj_star so i just need to think of an elegant way of keeping of keeping the monotonic int concurrent across each obj, or do i have a separate int for each object?

PR #1431, opened 2026-08-11
feat: zh-cn-localization

CaesinEachonson:
## Summary by cubic
Adds the first batch of Chinese localization for chapter creation and related UI, plus a language toggle in Settings. Extracts strings into `en.json`/`zh.json`, standardizes placeholders, applies CJK-safe fonts across UI (draw and measurement), and refreshes localized globals on language switch.

- New Features
  - Localized chapter selection, livery/role setup, role distribution toggles, homeworld/flagship flows, event texts, tooltips, and error/popups via `localize()`.
  - Centralized button labels with `localize_button_text()` and structured placeholders (`LANG_ENTRY_TEXT`/`LANG_ENTRY_VARIABLES`); UI text renderers, tooltips, and shutter buttons use `cjk_font()`; `faction_names` and stat rating labels rebuild from English sources via `refresh_locale_globals()`.

- Bug Fixes
  - Suppressed repeat missing-key warnings; safer placeholder replacement; normalized newline escapes across tooltips/inputs; ensured tooltip measurement and UI text rendering use CJK fonts.
  - Fixed crashes on language switch and when opening squad view (correct default unit focus); localized chapter load errors/popups.

<sup>Written for commit 7afe752d153ded550f09be2ae1ddd64ae633d34f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Adeptus-Dominus/ChapterMaster/pull/1431?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


scripts/scr_role_setup/scr_role_setup.gml
```diff
@@ -78,44 +78,44 @@ function update_role_data_wth_defaults() {
 /// @self Asset.GMObject.obj_creation
 function role_setup_objects() {
     specialist_distribution_box = new ToggleButton({
-        str1: "Equal Specialist Distribution",
+        str1: localize("Equal Specialist Distribution"),
```
OH296: as the toggle buttons are already a custom constructor we should not need to localise at call only on the update method of the ToggleButton this will save lots of extra ablative wrapping of text in localise calls

objects/obj_creation/Draw_0.gml
```diff
@@ -369,21 +369,21 @@ try {
...
         draw_set_halign(fa_right);
 
-        draw_text(650, 550, string_hash_to_newline("Imperium (" + string(disposition[2]) + ")"));
-        draw_text(650, 575, string_hash_to_newline("Adeptus Mechanicus (" + string(disposition[3]) + ")"));
-        draw_text(650, 600, string_hash_to_newline("Ecclesiarchy (" + string(disposition[5]) + ")"));
-        draw_text(650, 625, string_hash_to_newline("Inquisition (" + string(disposition[4]) + ")"));
+        draw_text(650, 550, string_hash_to_newline(localize("Imperium ({0})", [string(disposition[2])])));
+        draw_text(650, 575, string_hash_to_newline(localize("Adeptus Mechanicus ({0})", [string(disposition[3])])));
+        draw_text(650, 600, string_hash_to_newline(localize("Ecclesiarchy ({0})", [string(disposition[5])])));
+        draw_text(650, 625, string_hash_to_newline(localize("Inquisition ({0})", [string(disposition[4])])));
```
OH296: thees values (Imperium, Adeptus Mechanicus etc) sshould always be read from the global.faction_names and accessed using the enum eFACTION so global.faction_names[eFACTION.MECHANICUS] as result the faction_names global should get reset to the correct language values on language changes thus removing the need for localise calls where the translation is purely for the sake Faction names and other important variables

scripts/scr_role_setup/scr_role_setup.gml
```diff
@@ -78,44 +78,44 @@ function update_role_data_wth_defaults() {
...
     });
     scout_distribution_box = new ToggleButton({
-        str1: "Equal Scout Distribution",
+        str1: localize("Equal Scout Distribution"),
         font: fnt_40k_12,
         style: "box",
         x1: 710,
         y1: 250,
-        tooltip: $"Scout Distribution\nCheck if you wish for Scouts to be distributed equally across your Battle Companies rather than concentrated in the 10th.",
+        tooltip: localize("Scout Distribution\nCheck if you wish for Scouts to be distributed equally across your Battle Companies rather than concentrated in the 10th."),
```
OH296: same for tooltips, lets assume that all tooltips get localised as they're passed into the ToggleButton struct this saves going round the code base adding in a million wrapper calls

OH296: as a result this means that the update method i the ToggleButton constructor will need to account for the fact that localise varibles may be passed to the constructor, this could be done by allowing a seperate tooltip_variables key to be passed or by allowing tooltip to be a struct or an array that is sscanned for once it is being read in the update method of the constructor. Not sure what you think is best in this instance? 

either way a method that does not enforce changes to be made to ever single constrictor call as currently happens would be more ideal imo

scripts/scr_creation_draw_slides/scr_creation_draw_slides.gml
```diff
@@ -467,52 +467,52 @@ function draw_chapter_trait_select() {
...
+            localize("Uncooperative"),
+            localize("Uncooperative"),
+            localize("Uncooperative"),
+            localize("Neutral"),
+            localize("Trusted"),
+            localize("Trusted"),
+            localize("Trusted"),
+            localize("Trusted"),
+            localize("Exemplary"),
         ];
```
OH296, line 498: can we introduce a localise array funtionality (see my other comment about global.faction_names) this way instead of trying to localise each item one could just call localise_array(cooperation_ratings) and search for the array in the language pack, no need i think in this instance to make a default fallback either just assume english is the default fallback and search that file and if the value is not there bung and error message using LOGGER.error()

scripts/scr_creation_draw_slides/scr_creation_draw_slides.gml
```diff
@@ -437,7 +437,7 @@ function draw_chapter_trait_select() {
     }
 
     draw_set_color(CM_GREEN_COLOR);
-    draw_text_transformed(800, 120, string_hash_to_newline("Points: " + string(points) + "/" + string(maxpoints)), 0.6, 0.6, 0);
+    draw_text_transformed(800, 120, string_hash_to_newline(localize("Points: {0}/{1}", [string(points), string(maxpoints)])), 0.6, 0.6, 0);
```
OH296, line 440:
```suggestion
    draw_text_transformed(800, 120, string_hash_to_newline(localize("Points: {0}/{1}", [
points,maxpoints])), 0.6, 0.6, 0);
```

where posssible with interpolation remove the legacy string(<vari>) uses they are legacy and automatically handled by iterpolation

OH296:
Nice work extending the translation, I've made a few review points, bear in mind my word is not god if you think for some reason any of the points ii've made are advantages compared to choices you've made feel free to point that out and we can make an informed choice

scripts/scr_company_struct/scr_company_struct.gml
```diff
@@ -459,7 +459,7 @@ function CompanyStruct(comp) constructor {
     static default_member = function() {
         var _member = company_squads[0].fetch_member(0);
         if (is_struct(_member)) {
-            obj_controller.unit_focus = _fetched;
```
OH296, line 462: could you cherry pick this commit into a separate pr so it can be fast tracked

scripts/scr_buttons/scr_buttons.gml
```diff
@@ -202,7 +202,7 @@ function ReactiveString(text_param, x1_param = 0, y1_param = 0, data = {}) const
     text = text_param;
     font = fnt_40k_14;
```
OH296, line 203: can the cjk_font not simply be run here ad it thee update call so if the font is updated the correct new font is assigned using the cjk_font call this would reduce 90% of the cjk_font calls

scripts/scr_buttons/scr_buttons.gml
```diff
@@ -221,7 +221,7 @@ function ReactiveString(text_param, x1_param = 0, y1_param = 0, data = {}) const
     static update = function(data = {}) {
         move_data_to_current_scope(data);
         var temp_font = draw_get_font();
-        draw_set_font(font);
```
OH296, line 224: one way of doing the above point would be to create a sspecialised version of move_data_to_current_scope for mmoving data into constructorss that on reading the "font" key always runs  self.font = cjk_font(font);

scripts/scr_company_struct/scr_company_struct.gml
CaesinEachonson: It's not very safe i think. this one is to fix a crash on this PR.

scripts/scr_buttons/scr_buttons.gml
```diff
@@ -358,18 +377,33 @@ function LabeledIcon(icon_param, text_param, x1_param = 0, y1_param = 0, data =
...
+    ///       font_cjk current before calling.
+    /// @returns {undefined}
+    static measure_text_width = function() {
+        var _prev_font = draw_get_font();
+        draw_set_font(font_cjk);
+        text_width = string_width(text) + 2;
+        draw_set_font(_prev_font);
+    };
+
+    measure_text_width();
```
OH296: as update is run on the initialisation of all constructors and should be sole way data is updated within constructors so this iss not really needed just slam the logic straight into the update method it never needed to be callled separately in the first place

scripts/scr_buttons/scr_buttons.gml
```diff
@@ -1392,8 +1431,11 @@ function ToggleButton(data = {}) constructor {
 
     static update = function(data = {}) {
         move_data_to_current_scope(data);
+        str1 = localize_button_text(str1);
+        tooltip = localize_button_text(tooltip);
+        font_cjk = cjk_font(font);
```
OH296, line 1438: if font_cjk is just meant to be the defacto font there is no need to have two separate variables just use font = cjk_font(font); as you've done with the str1 variable and be done with it

scripts/scr_ChapterTraits/scr_ChapterTraits.gml
```diff
@@ -138,25 +138,25 @@ function ChapterTrait(trait) constructor {
                 var _line = "";
 
                 if (struct_exists(_mod, "int_mod") && _mod.int_mod != 0) {
-                    _line += $"  Disposition Gains : {string_plus_minus(_mod.int_mod)}{_mod.int_mod}\n";
+                    _line += localize("  Disposition Gains : {0}\n", [string_plus_minus(_mod.int_mod) + string(_mod.int_mod)]);
```
OH296: as per before unless calcs are being done always favour strong interpolation over string(<var>) methods
```suggestion
                    _line += localize("  Disposition Gains : {0}\n", [$"{string_plus_minus(_mod.int_mod)}{_mod.int_mod}]);
```

scripts/scr_buttons/scr_buttons.gml
```diff
@@ -220,8 +238,9 @@ function ReactiveString(text_param, x1_param = 0, y1_param = 0, data = {}) const
 
     static update = function(data = {}) {
         move_data_to_current_scope(data);
+        font_cjk = cjk_font(font);
```
OH296, line 256: agains this is overly verbose `font = cjk_font(font);` should be fine but if as per the first thought we kake move_data_to_current_scope_ui(data) and check for if "font" is one of the keys in the update packet then we can centralise all font = cjk_font(font); into that newly created function and apply it to all constructors same goes for sstr1 and tooltip localisation tests

OH296:
Okay very close now on this one just a few more tweaks and we should be good to go

scripts/scr_company_struct/scr_company_struct.gml
OH296: don't worry i have handled it

scripts/scr_buttons/scr_buttons.gml
CaesinEachonson: I cannot change it to font=cjk_font(font). It'll make the game unable to switch back to English once I changed language...

scripts/scr_buttons/scr_buttons.gml
```diff
@@ -358,18 +377,23 @@ function LabeledIcon(icon_param, text_param, x1_param = 0, y1_param = 0, data =
...
-    text_width = string_width(text) + 2;
-    draw_set_font(temp_font);
 
     static update = function(data = {}) {
         move_data_to_current_scope(data);
+        text = localize_button_text(text);
+        tooltip = localize_button_text(tooltip);
+        font_cjk = cjk_font(font);
         if (text_position == "right") {
+            var _prev_font = draw_get_font();
```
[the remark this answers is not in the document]
CaesinEachonson: I'll see if i can get this cubic thingdeployed on my computer so I don't have to spam committing next time...

scripts/scr_buttons/scr_buttons.gml
OH296: okay i get what you mean

OH296:
Okay i thinkk this is all ready sorrry t's taken me a few secs but iwill do a final scan and merge as soon as i have a moment

scripts/SettingsManager/SettingsManager.gml
```diff
@@ -75,6 +75,7 @@ function SettingsManager() constructor {
     static apply_language = function() {
```
The-Real-Nyx, line 75: This needs to refresh existing UI components. As at least the component update() methods I checked overwrite their source fields with the localized output
```
label = localize_button_text(label);
tooltip = localize_button_text(tooltip);
str1 = localize_button_text(str1);
```
So, after switching from english to chinese any field altered this way contains the chinese text and not the english which is fine except they have no way to restore the english text, leaving a user who switched from english to chinese and then back to english stuck with a bunch of leftover chinese text about.

Not saying to add that logic here but you need logic somewhere to ensure when a new language is applied everything gets updated to that language/restored back to the language.

Also just in general the way you currently have implemented the localization logic could not easily (to what I read of it) be extended for other localization's. The logic is broadly very explicitly for Chinese rather than abstracted from the particular language to localize. 

I am not nor can I tell you to make it extensible I just wanted to point it out because on the off chance anyone else wants to localize a language most of the logic (that I looked at least) would not be reusable and really for something like this you want it to be very plug and play with languages, separating the logic from the data files with the language so that the logic can easily plugin whatever language you feed it without requiring changes to itself per language.

CaesinEachonson: Yeah the UI refresh problem definitely need to be addressed with. I'll get it done later. The reason I overlooked it was because, well, the only place it can actually affect is the system names on the main map. That is the only part of the UI that won't be refreshed throughout the game.

As for other languages - I believe it won't be much of a problem. You can add a new language option then make another json to store all the strings from en.json to make translation. The font however, is indeed exclusive for Chinese (Will fix it later) as Chinese letters are not displayed with the game's original font. I'll see how to make it check which font to use based on which language is selected.

PR #1495, opened 2026-09-01
fix(inquisition): Restore Necron Tomb progression

The-Real-Nyx:
## Summary by cubic
Restores Necron Tomb mission progression after combat so the raid can reach completion.

- Extracts mission stage advancement into `advance_necron_tomb_mission` and reuses it after Necron combat encounters.
- Fixes the mission prompt queue so the popup opens at the front instead of appending to the current popup stack.
- Replaces direct `instance_destroy` calls with `popup_default_close` so optional popup exits close cleanly.
- Adds a post-battle summary line when the mission continues after combat.

<sup>Written for commit 46186d573c641c2537b20ab53c20dcf8642a0011. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Adeptus-Dominus/ChapterMaster/pull/1495?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/scr_inquisition_mission/scr_inquisition_mission.gml
+++ b/scripts/scr_inquisition_mission/scr_inquisition_mission.gml
@@ -717,6 +717,9 @@ function mission_investigate_planet() {
 }
 
 /// @self Asset.GMObject.obj_star
+/// @desc Queues the Necron Tomb mission prompt when a Plasma Bomb is present.
+/// @param {Real} planet Planet index containing the Necron Tomb.
+/// @returns {Undefined}
 function setup_necron_tomb_raid(planet) {
     LOGGER.info($"player on planet with necron mission {name} planet: {planet}");
     var have_bomb;
@@ -726,7 +729,7 @@ function setup_necron_tomb_raid(planet) {
         var tixt;
         tixt = $"Your marines on {planet_numeral_name(planet, id)}";
         tixt += " are prepared and ready to enter the Necron Tombs.  A Plasma Bomb is in tow.";
-        var _number = instance_exists(obj_turn_end) ? obj_turn_end.current_popup : 0;
+        var _number = instance_exists(obj_turn_end) ? 1 : 0;
         var _pop_data = {
             mission: "necron_tomb_excursion",
             loc: name,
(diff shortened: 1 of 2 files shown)


scripts/scr_inquisition_mission/scr_inquisition_mission.gml
```diff
@@ -726,7 +729,7 @@ function setup_necron_tomb_raid(planet) {
         var tixt;
         tixt = $"Your marines on {planet_numeral_name(planet, id)}";
         tixt += " are prepared and ready to enter the Necron Tombs.  A Plasma Bomb is in tow.";
-        var _number = instance_exists(obj_turn_end) ? obj_turn_end.current_popup : 0;
+        var _number = instance_exists(obj_turn_end) ? 1 : 0;
```
OH296, line 732: Think this is required to keep up with the popup sequence count when there are multiple stacked popups in a turn end

The-Real-Nyx: `_number` doesn't store the popup’s position in the sequence, it's used as a queue continuation flag inside `popup_default_close()`, which checks `if != 0` before scheduling `obj_turn_end.alarm[1]`. So, before this path would store 0, and then the recreated post combat popup would fail to rearm Alarm 1 when closed stalling the end turn popup queue, and preventing the player from ending their turn. And from testing it did for sure get stuck before sometimes and does for sure not get stuck now but if the change does somehow cause regression with the popup sequencing can you point out where that happens more specifically cause I don't see it.

EttyKitty:
was this tested? the description doesn't say.

The-Real-Nyx:
> was this tested? the description doesn't say.

Yeah. The bug report had a save attached so I was able to reproduce the bug prior to the fix and then test again post fix and verify it was fixed.

PR #1505, opened 2026-09-04
refactor: Update legacy code with UnitIndex

OH296:
---
## Summary by cubic
Replaces `obj_ncombat`'s legacy per-role counters and `big_mofo` gating with the `UnitIndex` system so unit tracking and battle narration come from a single source.

**Refactors**
- Battle text in `Alarm_1.gml` builds role summaries from `player_unit_index` using `role_count`, `plural_string_role`, and `sum_roles`; dreadnought counts include Venerable dreadnoughts.
- The battlecry speaker is chosen by a role priority list, with a named unit pulled from the index.
- Ship boarding and ship demon summon battles route through the standard `Roster`/`collect_role_group` path; the duplicate `scr_ship_battle` script is removed.
- `scr_roster` populates `player_unit_index` via `add_to_index`; legacy counter and `big_mofo` updates are dropped from `scr_roster` and `scr_civil_roster`.

**Migration**
- `big_mofo`, `important_dudes`, and the per-role counters are no longer maintained; anything referencing them must read from `player_unit_index` instead.

<sup>Written for commit 2c3e3b5997d127521eb4c226496aac149457016b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Adeptus-Dominus/ChapterMaster/pull/1505?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


EttyKitty (dismissed):

objects/obj_ncombat/Alarm_1.gml
```diff
@@ -32,10 +32,13 @@ var d7 = "";
...
-temp += techmarines + honors + dreadnoughts + terminators + captains;
-temp += standard_bearers + champions + important_dudes + chaplains + apothecaries;
-temp += sgts + vet_sgts;
+var _marine_roles = active_roles();
+array_push(_marine_roles, "Forge Master");
+array_push(_marine_roles, "Master of Sanctity");
+array_push(_marine_roles, "Master of the Apothecarion");
+array_push(_marine_roles, $"Chief {obj_ini.player_role_data[eROLE.LIBRARIAN].role}");
+
+var _marine_count = player_unit_index.sum_roles(_marine_roles);
```
The-Real-Nyx, line 37: `active_roles()` contains the base role names, so adding only the specialist heads still leaves out Codiciery, Lexicanum, Venerable Dreadnoughts, and specialist Aspirants, thus omits them from `_marine_count`.

```suggestion
var _marine_roles = active_roles();
array_push(_marine_roles, "Forge Master");
array_push(_marine_roles, "Master of Sanctity");
array_push(_marine_roles, "Master of the Apothecarion");
array_push(_marine_roles, $"Chief {obj_ini.player_role_data[eROLE.LIBRARIAN].role}");

var _extra_roles = array_concat(
    role_groups(SPECIALISTS_DREADNOUGHTS),
    role_groups(SPECIALISTS_LIBRARIANS, true),
    role_groups(SPECIALISTS_TRAINEES)
);

for (var i = 0; i < array_length(_extra_roles); i++) {
    var _extra_role = _extra_roles[i];
    if (!array_contains(_marine_roles, _extra_role)) {
        array_push(_marine_roles, _extra_role);
    }
}

var _marine_count = player_unit_index.sum_roles(_marine_roles);
```

objects/obj_ncombat/Alarm_1.gml
```diff
@@ -162,131 +167,44 @@ if (string_count("_attack", battle_special) > 0) {
...
+    for (var i = 0; i < array_length(_small_include); i++){
+        var _role = _small_include[i];
+        if (player_unit_index.role_count(_role) > 0){
+            p2 += player_unit_index.plural_string_role(_role) + ", ";
+        }
+    }
+} else {
+    var _others = _marine_count - player_unit_index.sum_roles(_auto_include);
+    _others -= _dread_count;
+    p2 += string(_others) + " other various Astartes, ";
```
The-Real-Nyx: The Chapter Master, Captains, Honour Guard, Ancients, Champions, and specialist heads contribute to `_marine_count`, but aren't included in the composition description for forces < 200 because the “other various Astartes” entry is only appended for forces > 200


```suggestion
p2 = "";

var _auto_include = [
    _marine_roles[eROLE.TACTICAL],
    _marine_roles[eROLE.VETERAN],
    _marine_roles[eROLE.ASSAULT],
    _marine_roles[eROLE.DEVASTATOR],
];

for (var i = 0; i < array_length(_auto_include); i++) {
    var _role = _auto_include[i];
    if (player_unit_index.role_count(_role) > 0) {
        p2 += player_unit_index.plural_string_role(_role) + ", ";
    }
}

var _small_include = [
    _marine_roles[eROLE.TERMINATOR],
    _marine_roles[eROLE.CHAPLAIN],
    _marine_roles[eROLE.APOTHECARY],
    _marine_roles[eROLE.LIBRARIAN],
    _marine_roles[eROLE.TECHMARINE],
    _marine_roles[eROLE.SERGEANT],
    _marine_roles[eROLE.VETERANSERGEANT],
    _marine_roles[eROLE.SCOUT],
];

var _other_count = _marine_count
    - player_unit_index.sum_roles(_auto_include)
    - _dread_count;

if (_marine_count < 200) {
    for (var i = 0; i < array_length(_small_include); i++) {
        var _role = _small_include[i];
        if (player_unit_index.role_count(_role) > 0) {
            p2 += player_unit_index.plural_string_role(_role) + ", ";
        }
    }

    _other_count -= player_unit_index.sum_roles(_small_include);
}

if (_other_count > 0) {
    p2 += $"{_other_count} other various Astartes, ";
```

objects/obj_ncombat/Alarm_1.gml
```diff
@@ -303,52 +221,34 @@ if (string_count(", ", p2) == 1) {
 }
 p2 += ".";
 
-if ((standard_bearers > 1) && (!dropping)) {
+if ((player_unit_index.role_count(_marine_roles[eROLE.ANCIENT]) > 1) && (!dropping)) {
     p5 = "  Chapter Ancients hold your Chapter heraldry high and proud.";
 }
 
-if (dreadnoughts + predators + land_raiders > 3) {
+if (_dread_count + predators + land_raiders > 3) {
```
The-Real-Nyx: If you deploy three or fewer dreadnoughts into battle with no other vehicles (like during a raid) they don't get counted in the display at all as they are excluded from other marines and excluded from the armoured division count as it would fail this gate.

objects/obj_ncombat/Alarm_1.gml
```diff
@@ -303,52 +221,34 @@ if (string_count(", ", p2) == 1) {
...
-        p6 += string(dreadnoughts) + " " + string(obj_ini.player_role_data[eROLE.DREADNOUGHT].role) + "s, ";
-    }
-
-    if (rhinos == 1) {
-        p6 += string(rhinos) + " Rhino, ";
-    }
-    if (rhinos > 1) {
-        p6 += string(rhinos) + " Rhinos, ";
+    if (_dread_count > 0) {
+        p6 += player_unit_index.plural_string_role(_marine_roles[eROLE.DREADNOUGHT]);
```
The-Real-Nyx: `_dread_count` includes ordinary and Venerable Dreadnoughts, but the current text counts only the ordinary role and the text is missing a trailing ", " which causes the cleanup to truncate the name or join it directly to the next vehicle entry.

```suggestion
if (_dread_count > 0) {
    p6 += string_plural_count(
        _marine_roles[eROLE.DREADNOUGHT],
        _dread_count,
        false
    ) + ", ";
```

The-Real-Nyx (commented):
I was kinda in rush so don't take my suggestions as absolutely correct or anything as I might've overlooked something.

OH296:
> I was kinda in rush so don't take my suggestions as absolutely correct or anything as I might've overlooked something.

All good, all the points are valid in one way or the other

objects/obj_ncombat/Alarm_1.gml
OH296: This follows what already existed, Would it be preferable for dreads to have their own text sequence?

objects/obj_ncombat/Alarm_1.gml
OH296: a currently enuming all roles and adding them to the player_role_data array to avoid having to do this

objects/obj_ncombat/Alarm_1.gml
The-Real-Nyx: > This follows what already existed, Would it be preferable for dreads to have their own text sequence?

Ah yeah you're right, that's my bad I probably glazed over looking at the diff. So, I guess it's just a preexisting bug then as they should presumably get counted somewhere in the combat display? So, giving them their own section would work.

objects/obj_ncombat/Alarm_1.gml
The-Real-Nyx: > a currently enuming all roles and adding them to the player_role_data array to avoid having to do this

Yeah, that is probably the better solution to the issue.

objects/obj_ncombat/Alarm_1.gml
OH296: ow section now added which ca be improved over time

EttyKitty (approved):
This is a code collapse. I'm not going to nitpick, because this is better than what was. A net positive.
If there are no bugs with this - it's fine. And I'm bad at spotting bugs with my eye on GH.

